PR #23, opened 2022-03-09
Reformulate nonlinear constraints correctly for CHR

hdavid16:
Manually replaces nonlinear constraints in model with changed constraints when applying convex hull reformulation.
Ensures that a model is given at least two constraints.


hdavid16:
fork needs to be rebased

PR #43, opened 2022-05-13
Data Structure Refactor

pulsipher:
This pull request will take some time and seeks to completely refactor DisjunctiveProgramming.jl to better extend JuMP in a more modular and performant way.

**TODO**
- [x] Finish basic mechanics for creating/storing disjunction constraints
- [x] Add types and methods to create/store logical propositions
- [x] Decide whether we should create our own logical variable type
- [ ] ~~Add printing~~ (save for a future PR)
- [x] Implement the reformulation techniques (Big-M, Hull, Indicator Constraints): affine, quadratic, and nonlinear 
- [x] Implement reformulations for logical propositions
- [x] Implement reformulations for logical selectors (exactly, at least, at most)  
- [x] Add macro(s) for disjunction addition
- [x] Add tests
- [ ] ~~Add documentation via Documenter.jl~~ (save for a future PR)
- [x] Support nested disjunctions
- [x] Update the README.md and move tutorial/guide to docs.


hdavid16:
TO DO:

- [ ] Implement methods to copy a GDP model: Need to copy disjunction constraints over to the new model (maybe something similar to what is done in copy_to in https://github.com/jump-dev/MathOptInterface.jl/blob/54f2c48715229e17e430d3f06c11d92f9d190f6c/src/Utilities/copy.jl). Otherwise, the constraints inside the disjuncts are not owned by the new model...

src/hull.jl
```diff
@@ -1,163 +1,74 @@
...
-            m[Symbol(var_i_ub)] = @constraint(m, m[var_name_i] .- UB * m[bin_var][i] .<= 0, base_name = var_i_ub)
+function _disaggregate_variables(model::JuMP.Model, disj::DisjunctiveConstraintData)
+    disj_vars = _get_variables(disj)
+    for d in disj.constraint.disjuncts
+        #create binary variable for logic variable (indicator)
+        bvar = JuMP.@variable(model, 
+            base_name = string(d.indicator), 
+            binary = true, 
+        )
+        model[Symbol(d.indicator,"_Bin")] = bvar
```
pulsipher, line 34: Rather than rely on the object dictionary which could have collisions and will make repeated transformations problematic, it would be better to augment `GDPData` to have a mapping dictionary that preserves this information i.e., mapping the disjunct to a variable like this.

This comment also applies to other places in the code that do the same thing, but I am not going to mark all of them.

src/hull.jl
```diff
@@ -1,163 +1,74 @@
...
+    end
+    for var in disj_vars
+        JuMP.is_binary(var) && continue #skip binary variables
+        sum_disag_vars = JuMP.AffExpr(0) #sum of disaggregated variables
+        for d in disj.constraint.disjuncts
+            #create disaggregated var
+            lb = min(0, JuMP.lower_bound(var))
+            ub = max(0, JuMP.upper_bound(var))
```
pulsipher, line 42: This may run into problems if the variables are unbounded

src/hull.jl
```diff
@@ -1,163 +1,74 @@
...
-            push!(dis_vars, var_i)
-        end
-        aggr_con = "$(var)_$(bin_var)_aggregation"
-        m[Symbol(aggr_con)] = @constraint(m, var == sum(dis_vars), base_name = aggr_con)
+function _disaggregated_constraint(model::JuMP.Model, con::JuMP.ScalarConstraint, lvar::LogicalVariableRef)
+    new_con_func = JuMP.AffExpr()
+    for (var, coeff) in con.func.terms
+        JuMP.is_binary(var) && continue #skip binary variables
+        disag_var = model[Symbol(var,"_",lvar)]
+        new_con_func.terms[disag_var] = coeff
```
pulsipher, line 70: Again this assumes the constraint is affine

src/propositions.jl
pulsipher, line 1: Most of this can be removed and refactored to leverage the new JuMP nonlinear interface

src/reformulate.jl
```diff
@@ -1,82 +1,229 @@
...
+    con_func.terms[bvar] = -con.set.value
+
+    JuMP.add_constraint(model,
+        JuMP.build_constraint(error,
+            con_func,
+            _MOI.EqualTo(0)
+        )
+    )
 end
-reformulate_constraint(args...) = nothing
```
pulsipher, line 82: It would be good to have a fallback for the `_reformulate` methods in case some other constraint type is intercepted.

pulsipher (commented):
This is great. My main recommendations pertain to not modifying the object dictionary, being more careful about constraint types, adding fallbacks, and adapting the code to leverage JuMP's new nonlinear interface.

src/reformulate.jl
hdavid16: Added these two definitions to /src/reformulate.jl: 
# define fallbacks for other constraint types
_reformulate(model::JuMP.Model, method::Hull, args...) = nothing
_reformulate(model::JuMP.Model, method::BigM, args...) = nothing

pulsipher: The problem with that is that such constraints will then be silently ignored. Better to throw an error if we are not going to reformulate them.

src/bigm.jl
```diff
@@ -1,40 +1,68 @@
...
-    set_normalized_coefficient(constr, bin_var_ref , M)
-end
-function big_m_reformulation!(constr::NonlinearConstraintRef, bin_var, M, i, j, k)
-    M = get_reform_param(M, i, j, k; constr)
-    #create symbolic variables (using Symbolics.jl)
-    for var_ref in constr.model[:gdp_variable_refs]
-        var_sym = Symbol(var_ref)
-        eval(:(Symbolics.@variables($var_sym)[1]))
+function _interval_arithmetic_GreaterThan(con::JuMP.ScalarConstraint, M::Float64)
+    for (var,coeff) in con.func.terms
```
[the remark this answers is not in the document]
hdavid16: fixed with the fallbacks on `_reformulate`

src/hull.jl
hdavid16: Added an error right before this:
```
if !JuMP.has_lower_bound(var) | !JuMP.has_upper_bound(var)
    error("Variable $var must have a lower and upper bound defined when using the Hull reformulation.")
end
```

src/hull.jl
hdavid16: added disaggregated_variables `Dict` to `GDPData`

pulsipher: Up to you, think about what would make your life easier with other function calls and functionality.

src/bigm.jl
```diff
@@ -1,40 +1,68 @@
...
+end
+_calculate_tight_M(con::JuMP.ScalarConstraint) = Inf
 
-    big_m_reformulation!(constr::NonlinearConstraintRef, bin_var, M, i, j, k)
+"""
 
-Perform Big-M reformulaiton on a nonlinear constraint at index k of constraint j in disjunct i.
+"""
+function _interval_arithmetic_LessThan(con::JuMP.ScalarConstraint, M::Float64)
+    for (var,coeff) in con.func.terms
```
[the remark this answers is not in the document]
hdavid16: Sounds, good. I have added the `AffExpr` part to it. We cannot use interval arithmetic with the variable bounds to find the maximum and minimum of a quadratic expression because nonlinear expressions need you to solve a max or min optimization problem to find the largest or smallest value the expression can take. For quadratic expressions we could take the derivative and find the zero if we wanted to support infering the tightest big M for them.

src/hull.jl
hdavid16: ok. I'll keep them separate for now. merging the two shouldn't be a problem down the road.

src/reformulate.jl
hdavid16: done. let me know if there is a better way to do this. I did:
```
function _reformulate(
    model::JuMP.Model, 
    method::AbstractReformulationMethod, 
    con::JuMP.AbstractConstraint, 
    lvar::LogicalVariableRef
)
    error("$method reformulation for constraint $con is not supported yet.")
end
```

src/bigm.jl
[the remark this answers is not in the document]
pulsipher: Ya for quadratics, I imagine there is an analytical formula we can code in.

src/reformulate.jl
pulsipher: That should do the trick

src/hull.jl
```diff
@@ -1,163 +1,74 @@
...
+    union([
+        _get_variables(con)
+        for con in d.constraints
+    ]...)
+end
+function _get_variables(con::JuMP.AbstractArray{<:JuMP.AbstractConstraint})
+    union(_get_variables.(c)...)
+end
+function _get_variables(con::JuMP.ScalarConstraint{JuMP.AffExpr, T}) where {T}
+    collect(keys(con.func.terms))
```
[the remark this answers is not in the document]
hdavid16: It looks like you're doing the same thing there for Affine Expressions. I added the one for Quadratic Constraints. What do you mean about more efficient? Are you referring to my use of `union`? I do this to remove duplicates when I look at all the constraints in an array of constraints, in a disjunct or in a disjunction.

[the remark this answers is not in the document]
pulsipher: Notice my use of `Set` instead of an array. This ends up being more efficient than using union with Arrays.

src/constraints.jl
```diff
@@ -0,0 +1,645 @@
...
+    lhs, 
+    rhs
+)
+    op = val_op == Val{:⇒} ? :⇒ :
+            val_op == Val{:⇔} ? :⇔ : :(<-->)
+    error_msg = 
+        "The MOI set in the logical constraint $lhs $op $rhs was not identified. " *
+        "The set `MOI.EqualTo{Bool}(value)` should be preceded by `==` or `in`."
+    if rhs.args[1] in (:in, :(==))
+        rhs0 = rhs.args[2]
```
pulsipher, line 459: A few things here:
- You need to use `_error` instead of `error` to get a nice macro error message.
- The accepted syntax should only be `[logical_expr] in set` where `set` is a `MOI.EqualTo` or `[logical_expr] == true` (or `false`). However, it currently can mix the `in` and `==` operators between these two syntaxes.
- The user might also use `∈` instead of `in`.
- You can only check the type of AST arguments if they are explicit numbers or booleans, otherwise they cannot be determined until runtime. Thus, the  `rhs.args[3] isa` checks will not work as you intend. In reality the acceptable forms of `rhs.args[3]` could include `Bool`, `Symbol` (some Julia variable was given that points to a set or boolean), or `Expr` where the head type is `call` and the first argument is `:(MOI.EqualTo)`. To properly handle the dispatch, you'll need some some function to put in the parse or build code that will do appropriate things at runtime (i.e., when the code is executed).

src/constraints.jl
```diff
@@ -0,0 +1,645 @@
...
+either be true or false.
+"""
+function JuMP.build_constraint(
+    _error::Function, 
+    func::AbstractVector{<:Union{Number, LogicalVariableRef, _LogicalExpr}}, # allow any vector-like JuMP container
+    set::_MOI.AbstractVectorSet # Keep general to allow CP sets from MOI
+    )
+    return JuMP.VectorConstraint(func, set)
+end
+
```
pulsipher, line 546: We'll need to add dispatch methods for `_MOI.EqualTo` (not `Bool`) to check and convert to `Bool` where possible and then call the other `build_constraint` methods.

hdavid16: done

src/reformulate.jl
```diff
@@ -1,82 +1,387 @@
...
+
+"""
+function _reformulate_disjunct(model::JuMP.Model, d::Disjunct, method::AbstractReformulationMethod)
+    #reformulate each constraint and add to the model
+    ind_idx = JuMP.index(d.indicator) #logical indicator
+    bv_idx = _indicator_to_binary(model)[ind_idx]
+    bvref = JuMP.VariableRef(model, bv_idx)
+    for con in d.constraints
+        cdata = _disjunct_constraints(model)[JuMP.index(con)]
+        _reformulate_disjunctive_constraint(model, cdata.constraint, bvref, method)
```
pulsipher, line 56: Is this method needed or used?

src/constraints.jl
```diff
@@ -0,0 +1,738 @@
...
+        "should be preceded by `in` or `∈`. Otherwise `func == Bool` should be used " *
+        "to define the constraint."
+    rhs0, set0 = rhs.args[2:3]
+    if rhs.args[1] in (:in, :∈) && set0.head == :call && set0.args[2] isa Bool    
+        set = set0
+    elseif rhs.args[1] == :(==) && set0 isa Bool
+        set = _MOI.EqualTo(set0)
+    else
+        _error(error_msg)
+    end
```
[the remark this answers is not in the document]
pulsipher: The proper way to do this would be to just give the raw `set0` to a function in the build code that will properly check it at runtime.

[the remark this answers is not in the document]
pulsipher: So you would have something like `code = :( build_constraint(_error, func, _process_raw_set(_error, set0)) )` where `_process_raw_set` is a collection of overloaded functions to deal with the set via dispatch as you want to.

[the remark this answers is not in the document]
pulsipher: I see that `<-->` is already built into JuMP, so we cannot touch that anyways.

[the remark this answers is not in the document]
pulsipher: And if we do put just an error, a safer place to do so would be with `operator_to_set`.

src/reformulate.jl
hdavid16: Yes, it is called from `_reformulate_disjuncts`. The reformulation is made on each disjunct of the disjunction.

src/reformulate.jl
```diff
@@ -1,82 +1,387 @@
...
 
-    # return m[bin_var]
+"""
+function _reformulate_disjunctive_constraint(
+    model::JuMP.Model,
+    con::JuMP.ScalarConstraint{T, S}, 
+    bvref::JuMP.VariableRef,
+    method::BigM
+) where {T, S <: _MOI.LessThan}
+    #TODO: need to pass _error to build_constraint
```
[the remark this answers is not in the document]
hdavid16: ok. That comment was just from copy and pasting code.

src/reformulate.jl
pulsipher: Right, but is this particular dispatch used with `AbstractReformulationMethod`? It seems you end up having to make a dispatch for each particular method.

src/constraints.jl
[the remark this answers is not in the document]
pulsipher: The `parse_code` is simply executed immediately before the `build_code`, but again this detail becomes uncesessary if we choose not to extend `parse_constraint_call`.

In JuMP, if we use `<--->` this happens:
```julia
@constraint(m, y1 <--> y2 == true)
ERROR: LoadError: At REPL[12]:1: `@constraint(m, y1 <--> y2 == true)`: Invalid right-hand side `y2 == true` of reified constraint. Expected constraint surrounded by `{` and `}`.
```
Thus, I would suggest we forego the use of both `-->` and `<-->` to avoid any confusion.

I saw the recursive strategy, we could explore something like this, but again there are a lot of edge cases to consider. And accounting for vectorized syntax compounds the complexity further. I think the problem is that we are working against the intended design of `parse_constraint_call` which is intended to enable constraints of the form `@constraint(m, myfunc(x, y))`, not intercept specifics of the constraint LHS to change the order of operations.

For this, my opinion would be to not extend `parse_constraint_call` and leave the default error message for now. If people complain, we could then extend `operator_in_set` for `⇔` and `⇒` to throw a better error. The consequence to this is that is would prevent other extensions from creating constraints using these operators to describe some new set type.

src/reformulate.jl
hdavid16: Yes you dispatch for each method at the disjunct constraint level

pulsipher:
We can use `operator_to_set` which gets called by one of the fallbacks fot
`parse_constraint_call`.

src/reformulate.jl
```diff
@@ -1,82 +1,387 @@
...
+        build_constraint(error, JuMP.NonlinearExpr(:-, Any[JuMP.NonlinearExpr(:+, bvrefs), bvref]), _MOI.LessThan(0))
+    )
+end
+function _reformulate_selector(model::JuMP.Model, ::MOIExactly, lvref::LogicalVariableRef, lvrefs::Vector{LogicalVariableRef})
+    bvref = _indicator_to_binary_ref(lvref)
+    bvrefs = Vector{Any}(_indicator_to_binary_ref.(lvrefs))
+    return JuMP.add_constraint(model,
+        build_constraint(error, JuMP.NonlinearExpr(:-, Any[JuMP.NonlinearExpr(:+, bvrefs), bvref]), _MOI.EqualTo(0))
+    )
+end
```
pulsipher, line 335: Are these methods really needed? Having a logic variable for `n` doesn't seem to have any practical use.

pulsipher:
I'll also leave a note that the expression creation in the reformulation steps can definitely be improved. Simply using `@expression` where possible would make a difference.

src/reformulate.jl
```diff
@@ -1,82 +1,387 @@
...
-Reformulate disjunction on a JuMP model.
+Create binary (indicator) variables for logic variables.
+"""
+function _reformulate_logical_variables(model::JuMP.Model)
+    for (lv_idx, lv_data) in _logical_variables(model)
+        var = JuMP.ScalarVariable(_variable_info(binary=true))
+        bvref = JuMP.add_variable(model, var, lv_data.name)
+        _indicator_to_binary(model)[lv_idx] = JuMP.index(bvref)
+    end
+end
```
pulsipher, line 14: TODO: add start value from the logical variable

src/reformulate.jl
hdavid16: It does for Nested relationships where the enforcement of the logical constraint depends on the parent.

pulsipher: Ok, we'll want to refactor these to create affine constraints then instead of nonlinear ones.

hdavid16: done

src/reformulate.jl
hdavid16: done

src/datatypes.jl
```diff
@@ -367,19 +365,17 @@ mutable struct GDPData
...
-    
-    # Map of disaggregated variables 
-    global_to_disjunct_variable::Dict{Tuple{_MOI.VariableIndex,_MOI.VariableIndex}, _MOI.VariableIndex}
-    global_to_disjunction_variables::Dict{Tuple{_MOI.VariableIndex,DisjunctionIndex}, Vector{_MOI.VariableIndex}}
 
     # Map of variable bounds
     variable_bounds::Dict{_MOI.VariableIndex, Tuple{Float64, Float64}} # TODO allow for other precision
 
+    # Reformulation variables and constraints
+    reformulation_variables::Set{_MOI.VariableIndex}
```
pulsipher, line 376: Using a `Vector` will probably be more efficient.

hdavid16: ok, changed

src/variables.jl
```diff
@@ -0,0 +1,259 @@
...
+    fix_value::Union{Nothing, Bool}=nothing
+)
+    JuMP.VariableInfo(
+        !isinf(lower_bound), lower_bound, 
+        !isinf(upper_bound), upper_bound, 
+        !isnothing(fix_value), fix_value, #fix value
+        !isnothing(start_value), start_value, #start value
+        binary, false #integrality
+    )
+end
```
pulsipher, line 18: I know I said differently before, but I think we should remove this and just create the reformulation variables using `@variable`.

src/variables.jl
```diff
@@ -0,0 +1,259 @@
...
+    data.logical_variables[JuMP.index(vref)].variable = new_var
+    return
+end
+function JuMP.unfix(vref::LogicalVariableRef)
+    data = gdp_data(JuMP.owner_model(vref))
+    var = data.logical_variables[JuMP.index(vref)].variable
+    new_var = LogicalVariable(nothing, var.start_value)
+    data.logical_variables[JuMP.index(vref)].variable = new_var
+    return
+end
```
pulsipher, line 134: We should document these with docstrings to include in the docs.

pulsipher: Also some of these should use `_set_ready_to_optimize(model, false)`

src/constraints.jl
```diff
@@ -0,0 +1,533 @@
...
+    _error::Function, 
+    func::Vector{T}, 
+    set::Union{_MOI.LessThan{Bool}, _MOI.GreaterThan{Bool}, _MOI.EqualTo{Bool}},
+    tag::DisjunctConstraint
+) where {T <: JuMP.AbstractJuMPScalar}
+    dim = length(func)
+    mapped_set = _scalar_to_vec_set(set,dim)
+    constr = JuMP.build_constraint(_error, func, mapped_set)
+    return _DisjunctConstraint(constr, tag.indicator)
+end
```
pulsipher, line 187: This approach will likely have unintended behavior and likely to break with future JuMP feature additions. It would be simpler and more stable to instead define:
```julia
function JuMP.build_constraint(
    _error::Function, 
    func, 
    set::JuMP.Nonnegatives, 
    tag::DisjunctConstraint
)
    constr = JuMP.build_constraint(_error, func, set)
    return _DisjunctConstraint(constr, tag.indicator)
end
```
for each of the three JuMP sets `Nonnegatives`, `Nonpositives`, and `Zeros`.

src/constraints.jl
```diff
@@ -0,0 +1,533 @@
...
+            _error("`$lvref` is not a valid logical variable reference.")
+        elseif !haskey(_indicator_to_constraints(model), JuMP.index(lvref))
+            _error("`$lvref` is not associated with any constraints.")
+        end
+    end
+    return s
+end
+
+# TODO account for nested disjunction inputs with indicators
+# TODO maybe replace Vectors for Tuple structured input to avoid nuance mistakes
```
pulsipher, line 258: These can just be removed now

src/constraints.jl
```diff
@@ -0,0 +1,533 @@
...
+        func::_LogicalExpr,
+        set::_MOI.EqualTo{Bool}
+    )
+
+Extend `JuMP.build_constraint` to add logical propositional constraints to a [`GDPModel`](@ref). 
+This in combination with `JuMP.add_constraint` enables the use of 
+`@constraint(model, [name], logical_expr == true/false)` to define a Boolean expression that must
+either be true or false.
+"""
+# Cardinality logical constraint
```
pulsipher, line 397: This comment needs to be removed in order for the docstring to register with the function

src/optimize.jl
```diff
@@ -0,0 +1,36 @@
...
+    end
+    gdp_data(model).reformulation_variables = Vector{_MOI.VariableIndex}()
+    #reformulate
+    _reformulate_logical_variables(model)
+    _reformulate_disjunctions(model, method)
+    _reformulate_logical_constraints(model)
+    #set solution method
+    _set_solution_method(model, method)
+    _set_ready_to_optimize(model, true)
+end
```
pulsipher, line 36: This should be refactored such that `_reformulate_disjunctions` is a public methods that serves as an extension point in our API for new reformulation methods.

src/variables.jl
hdavid16: done

src/constraints.jl
hdavid16: I agree, but the problem is that if we do this, the method is never called because these constraints will first go through JuMP's method:
```
function build_constraint(
    _error::Function,
    f,
    set::Nonpositives,
    args...;
    kwargs...,
)
    return build_constraint(_error, f, MOI.LessThan(false), args...; kwargs...)
end
```

hdavid16: nevermind, your suggestion works

src/optimize.jl
hdavid16: Ok. Should we still make `reformulate_model` public in case somebody wants to reformulate before solve?

pulsipher: Ya, we should. For instance, that is what David wanted.

pulsipher: Also for the reformulation storage vectors, we can just use `empty!` instead defining new ones each time.

src/variables.jl
hdavid16: done

src/optimize.jl
hdavid16: I think it would make more sense to make `_reformulate_disjunct_constraint` public so that it can be extended for other methods.

pulsipher: After taking a closer look, I think `_reformulate_disjunction` would make the most sense in case the user wants to do a  method that requires looking at the entire disjunction and not just an individual constraint from it.

src/constraints.jl
```diff
@@ -0,0 +1,559 @@
...
+end
+function _scalar_to_vec_set(set::_MOI.EqualTo{Bool}, dim::Int)
+    @assert _set_value(set) == false "Error parsing constraint set."
+    return _MOI.Zeros(dim)
+end
+
+# helper functions to reformulate vector constraints to indicator constraints
+_vec_to_scalar_set(set::_MOI.Nonpositives) = _MOI.LessThan(0)
+_vec_to_scalar_set(set::_MOI.Nonnegatives) = _MOI.GreaterThan(0)
+_vec_to_scalar_set(set::_MOI.Zeros) = _MOI.EqualTo(0)
```
pulsipher, line 28: I think we should delete these.

src/constraints.jl
```diff
@@ -0,0 +1,559 @@
...
+Delete a disjunction constraint from the `GDP model`.
+"""
+function JuMP.delete(model::JuMP.Model, cref::DisjunctionRef)
+    @assert JuMP.is_valid(model, cref) "Disjunctive constraint does not belong to model."
+    cidx = JuMP.index(cref)
+    dict = _disjunctions(model)
+    #delete each logical variable
+    for lv_idx in dict[cidx].constraint.disjuncts
+        JuMP.delete(model, LogicalVariableRef(model, lv_idx))
+    end
```
pulsipher, line 124: Since we are forcing users to create logical variables and associate them with disjunction constraints, wouldn't deleting them here invalidate the constraints that aren't deleted?

src/constraints.jl
```diff
@@ -0,0 +1,559 @@
...
+"""
+function JuMP.delete(model::JuMP.Model, cref::DisjunctConstraintRef)
+    @assert JuMP.is_valid(model, cref) "Disjunctive constraint does not belong to model."
+    cidx = JuMP.index(cref)
+    dict = _disjunct_constraints(model)
+    #delete from gdp_data
+    delete!(dict, cidx)
+    #delete from constraint_to_indicator mapping
+    delete!(_constraint_to_indicator(model), cidx)
+    # TODO: delete reformulated constraint? This would require mapping cidx to reformulated constraint index
```
pulsipher, line 145: I don't think we need to since it sets `_set_ready_to_optimize(model, false)`.

src/constraints.jl
```diff
@@ -0,0 +1,559 @@
...
+function _disjunction(
+    _error::Function,
+    model::JuMP.Model, # TODO: generalize to AbstractModel
+    structure,
+    name::String,
+    tag::DisjunctConstraint
+)
+    dref = _disjunction(_error, model, structure, name)
+    obj = JuMP.constraint_object(dref)
+    return JuMP.add_constraint(model, _DisjunctConstraint(obj, tag.indicator), name)
```
pulsipher, line 333: Are we ok with storing nested disjunctions twice?

src/constraints.jl
```diff
@@ -0,0 +1,559 @@
...
+    name::String = ""; 
+    nested::Bool = false,
+    indicator::LogicalVariableRef
+) # TODO add kw argument to build exactly 1 constraint
+    if nested
+        return _disjunction(error, model, structure, name, DisjunctConstraint)
+    else
+        return _disjunction(error, model, structure, name, DisjunctConstraint(indicator))
+    end
+end
```
pulsipher, line 372: This has a few problems. One is that regardless of `nested` it will always add the `DisjunctConstraint` which is incorrectly specified. Moreover, the indicator keyword argument doesn't have a default value. I would recommend the following two definitions:
```julia
function disjunction(
    model::JuMP.Model, 
    disjunct_indicators, # the docstring should have Vector{LogicalVariableRef} even though we don't explicitly type it this way
    name::String = ""
) # TODO add kw argument to build exactly 1 constraint
    return _disjunction(error, model, structure, name)
end

function disjunction(
    model::JuMP.Model, 
    disjunct_indicators, # the docstring should have Vector{LogicalVariableRef} even though we don't explicitly type it this way
    nested_tag::DisjunctConstraint
    name::String = ""
) # TODO add kw argument to build exactly 1 constraint
    return _disjunction(error, model, structure, name, nested_tag)
end
```

src/datatypes.jl
```diff
@@ -0,0 +1,406 @@
...
+    _MOIExactly{T<:Union{Int,LogicalVariableRef}} <: _MOI.AbstractVectorSet
+
+MOI level set for Exactly constraints, see [`Exactly`](@ref) for recommended syntax.
+"""
+struct MOIExactly{T<:Union{Int,LogicalVariableRef}} <: _MOI.AbstractVectorSet
+    value::T 
+    dimension::Int
+end
+
+const MOISelector = Union{MOIAtLeast, MOIAtMost, MOIExactly}
```
pulsipher, line 89: We should make this an abstract type that the sets inherit. So:
```julia
# TODO add docstring
abstract type MOISelector <: _MOI.AbstractVectorSet

struct MOIExactly{T<:Union{Int,LogicalVariableRef}} <: MOISelector
    value::T 
    dimension::Int
end
```

src/datatypes.jl
```diff
@@ -0,0 +1,406 @@
...
+
+A type for a disjunctive constraint that is comprised of a collection of 
+disjuncts of indicated by a unique [`LogicalVariableRef`](@ref).
+
+**Fields**
+- `disjuncts::Vector{LogicalVariableIndex}`: The disjuncts that comprise the constraint.
+"""
+struct Disjunction <: JuMP.AbstractConstraint
+    disjuncts::Vector{LogicalVariableIndex}
+end
```
pulsipher, line 229: Should we rename `disjuncts` to `indicators` or `disjunct_indicators` to be more clear?

src/reformulate.jl
```diff
@@ -1,82 +1,454 @@
...
+function _reformulate_disjunct(model::JuMP.Model, ind_idx::LogicalVariableIndex, method::AbstractReformulationMethod)
+    #reformulate each constraint and add to the model
+    bv_idx = _indicator_to_binary(model)[ind_idx]
+    bvref = JuMP.VariableRef(model, bv_idx)
+    for cidx in _indicator_to_constraints(model)[ind_idx]
+        cdata = _disjunct_constraints(model)[cidx]
+        reformulate_disjunct_constraint(model, cdata.constraint, bvref, method)
     end
-    return new_constraints
+end
```
pulsipher, line 58: We should add an extension point with a docstring to explain the API:
```julia
"""
my docstring
"""
function reformulate_disjunct_constraint end
```
I also think we should make `_reformulate_disjunction` public as well.

src/constraints.jl
hdavid16: Are you referring to deleting the reformulated constraint (TODO comment)? Or to deleting the `constraint_to_indicator` mapping?

src/constraints.jl
hdavid16: I haven't looked into this code on how nesting will work. I'll play around a bit with it. However, we might want to leave nesting for a separate PR. What are your thoughts?

src/constraints.jl
hdavid16: I don't think these functions are being used anywhere...

src/datatypes.jl
hdavid16: I like indicators

src/constraints.jl
pulsipher: Ok that  makes sense

src/constraints.jl
pulsipher: I am referring to the TODO. And yes it would make sense to not bother deleting the reformulated variables either.

src/constraints.jl
pulsipher: My thoughts are that it would be best to go ahead and support nesting since doing so might require breaking changes. The current definition structure should work. The only caveat is we store it twice currently and I am not sure if the reformulations are correct with it.

src/constraints.jl
pulsipher: They are intended for users to have a non macro option to define a single disjunction.

src/constraints.jl
pulsipher: I am guessing we need a way of knowing if a disjunction is nested or not for reformulating

src/reformulate.jl
```diff
@@ -1,82 +1,454 @@
...
+function reformulate_disjunct_constraint(
+    model::JuMP.Model,
+    con::JuMP.VectorConstraint{T, S, R}, 
+    bvref::JuMP.VariableRef,
+    method::BigM
+) where {T, S <: _MOI.Nonpositives, R}
+    M = JuMP.@expression(model, [i=1:con.set.dimension],
+        _get_M_value(method, con.func[i], con.set)
+    )
```
pulsipher, line 85: We should just use a comprehension here instead of `@expression` since this is just making a vecotr of constants.

src/reformulate.jl
```diff
@@ -1,82 +1,454 @@
...
+end
+function reformulate_disjunct_constraint(
+    model::JuMP.Model, 
+    con::JuMP.VectorConstraint{T, S, R}, 
+    bvref::JuMP.VariableRef,
+    method::BigM
+) where {T, S <: _MOI.Zeros, R}
+    M = JuMP.@expression(model, [i=1:con.set.dimension],
+        _get_M_value(method, con.func[i], con.set)
+    )
```
pulsipher, line 153: Again we should use a vector comprehension instead

src/constraints.jl
hdavid16: done

pulsipher:
My last suggestion would be to move all the big-m specific methods to bigm.jl and all the hull specific methods to hull.jl

src/reformulate.jl
hdavid16: Why is that preferred?

pulsipher: Well, we aren't generating any expressions here. So all the extra overhead from `@expression` is not needed and at first glance it seems to imply `M` will be a vector of JuMP expressions. More compactly we can just write:
```julia
M = [_get_M_value(method, con.func[i], con.set) for i in 1:con.set.dimension]
```

hdavid16: done

src/constraints.jl
hdavid16: Yes, we need to be able to reformulate nested disjunctions first so that we can add the reformulation as a disjunct constraint(s) and then reformulate the upper level disjunctions...

hdavid16: It get's tricky once we have multiple levels because we have to work our way from the inner most disjunction. Almost sounds like we should have some object tree tracking the nested relationship (dependency graph)...

pulsipher: Ya this is going to need some careful thought. I would be hesitant to add to disjunct constraints while reformulating since that will make it tricky to change to anothet reformulation. We can maybe make some buffer field to store them as we build the reformulation. Also note that th e disjunctions are stored in creation order and the inner most must be created first so they will be reformulated first.

hdavid16:
That’s a good point.
There is no guarantee that they will be created from the inside out since
all you need is to specify the logical variables for each disjunction.

src/constraints.jl
pulsipher: After thinking about it more, I think we can do the reformulations from the top down via recursion which will in turn evaluate it from the inside out. We should just store whether a disjunction is nested or not. If it isn't then reformulate it and it's underlying disjunctions otherwise skip it. We can set up a call to discuss this further if you want

hdavid16: done

src/constraints.jl
```diff
@@ -0,0 +1,555 @@
...
+    _error::Function,
+    model::JuMP.Model, # TODO: generalize to AbstractModel
+    structure,
+    name::String,
+    tag::DisjunctConstraint
+)
+    dref = _create_disjunction(_error, model, structure, name, true)
+    obj = JuMP.constraint_object(dref)
+    _add_indicator_var(_DisjunctConstraint(obj, tag.indicator), dref, model)
+end
```
pulsipher, line 321: This doesn't return the disjunction reference

src/constraints.jl
```diff
@@ -0,0 +1,555 @@
...
+
+# Disjunction build for unnested disjunctions
+function _disjunction(
+    _error::Function,
+    model::JuMP.Model, # TODO: generalize to AbstractModel
+    structure::Vector,
+    name::String
+)
+    _create_disjunction(_error, model, structure, name, false)
```
pulsipher, line 297: This should use `return` as per the style guide
```suggestion
    return _create_disjunction(_error, model, structure, name, false)
```

src/reformulate.jl
```diff
@@ -1,82 +1,168 @@
...
-        sum_disaggregated_variables(m, disj, bin_var)
+################################################################################
+#                              LOGICAL VARIABLES
+################################################################################
+# create binary (indicator) variables for logic variables.
+function _reformulate_logical_variables(model::JuMP.Model)
+    for (lv_idx, lv_data) in _logical_variables(model)
+        lv = lv_data.variable
+        lvref = LogicalVariableRef(model, lv_idx)
+        bvref = JuMP.@variable(model, base_name = lv_data.name, binary=true, start = lv.start_value)
```
pulsipher, line 37: Does this work if `lv.start_value = nothing`?

src/reformulate.jl
```diff
@@ -1,82 +1,168 @@
...
-    reformulate_constraint(constr::Tuple, bin_var, reformulation, param, i)
+    reformulate_disjunction(
+        model::JuMP.Model, 
+        disj::Disjunction,
+        method::AbstractReformulationMethod
+    ) where {T<:Disjunction}
 
-Reformulate a Tuple of constraints.
+Reformulate a disjunction using the specified `method`. Current reformulation methods include
+`BigM`, `Hull`, and `Indicator`. This method can be extended for other reformualtion techniques.
```
pulsipher, line 78:
```suggestion
`BigM`, `Hull`, and `Indicator`. This method can be extended for other reformulation techniques.
```

src/reformulate.jl
```diff
@@ -1,82 +1,168 @@
...
-    reformulate_constraint(constr::AbstractArray{<:ConstraintRef}, bin_var, reformulation, param, i, j = missing)
+The `disj` field is the `ConstraintData` object for the disjunction, stored in the 
+`disjunctions` field of the `GDPData` object.
+"""
+# generic fallback (e.g., BigM, Indicator)
+function reformulate_disjunction(model::JuMP.Model, disj::Disjunction, method::AbstractReformulationMethod)
+    ref_cons = Vector{JuMP.AbstractConstraint}()
+    for d in disj.indicators
+        push!(ref_cons, _reformulate_disjunct(model, d, method)...)
+    end
```
pulsipher, line 89: We should remove this space

src/reformulate.jl
hdavid16: Yes, `nothing` is the default start value on a logical variable

src/reformulate.jl
hdavid16: the space above the `return` statement?

pulsipher: yes

src/reformulate.jl
```diff
@@ -1,82 +1,168 @@
...
+    for vref in disj_vrefs #create sum constraint for disaggregated variables
+        push!(ref_cons, _aggregate_variable(model, vref, hull))
+    end
 
-    reformulate_constraint(constr::ConstraintRef, bin_var, reformulation, param, i, j = missing, k = missing)
+    return ref_cons
+end
+function reformulate_disjunction(model::JuMP.Model, disj::Disjunction, method::_Hull)
+    reformulate_disjunction(model, disj, Hull(method.value, method.variable_bounds))
```
pulsipher, line 108: Add `return`

src/reformulate.jl
```diff
@@ -1,82 +1,168 @@
...
+    #reformulate each constraint and add to the model
+    bvref = _indicator_to_binary(model)[lvref]
+    ref_cons = Vector{JuMP.AbstractConstraint}()
+    for cref in _indicator_to_constraints(model)[lvref]
+        con = JuMP.constraint_object(cref)
+        push!(ref_cons, reformulate_disjunct_constraint(model, con, bvref, method)...)
     end
+
+    return ref_cons
 end
```
pulsipher, line 122: Remove the space and use `append!`

src/indicator.jl
```diff
@@ -0,0 +1,40 @@
...
+    bvref::JuMP.VariableRef,
+    method::Indicator
+) where {T, S}
+    set = _vec_to_scalar_set(con.set)
+    reform_cons = Vector{JuMP.VectorConstraint}()
+    for (i, func) in enumerate(con.func)
+        reform_con = JuMP.build_constraint(error, [1*bvref, func], _MOI.Indicator{_MOI.ACTIVATE_ON_ONE}(set))
+        push!(reform_cons, reform_con)
+    end
+
```
pulsipher, line 28:
```suggestion
    return [JuMP.build_constraint(error, [1*bvref, f], _MOI.Indicator{_MOI.ACTIVATE_ON_ONE}(set)) for f in con.func]
```

src/indicator.jl
```diff
@@ -0,0 +1,40 @@
...
+    return reform_cons
```
pulsipher, line 29:
```suggestion
```

pulsipher:
Also, let's remove the extra spaces above most of the `return`s

src/hull.jl
```diff
@@ -1,163 +1,246 @@
...
+    else
+        lb = min(0, JuMP.lower_bound(vref))
+        ub = max(0, JuMP.upper_bound(vref))
+    end
+    return lb, ub
+end
+function _disaggregate_variables(model::JuMP.Model, lvref::LogicalVariableRef, vrefs::Set{JuMP.VariableRef}, method::_Hull)
+    #create disaggregated variables for that disjunct
+    for vref in vrefs
+        JuMP.is_binary(vref) && continue #skip binary variables
```
pulsipher, line 24: What about integer variables?

src/hull.jl
```diff
@@ -1,163 +1,246 @@
...
+    #create bounding constraints
+    dvname = JuMP.name(dvref)
+    new_con_lb_ref = JuMP.add_constraint(model,
+        JuMP.build_constraint(error, lb*bvref - dvref, _MOI.LessThan(0)),
+        "$dvname lower bounding"
+    )
+    new_con_ub_ref = JuMP.add_constraint(model,
+        JuMP.build_constraint(error, dvref - ub*bvref, _MOI.LessThan(0)),
+        "$dvname upper bounding"
     )
```
pulsipher, line 50: Names should not have white spaces, it messes up the LaTeX.

src/reformulate.jl
hdavid16: done for all functions

src/hull.jl
hdavid16: GDP currently assumes all variables are continuous. Integer variables should never show up in GDP, although there is some work by John Hooker where they use mixed-integer constraints inside of the disjuncts. We only skip the binary variables to accommodate for when we have nested disjunctions since the inner reformulated constraints will have binary variables.

src/indicator.jl
hdavid16: done

src/hull.jl
pulsipher: We should probably have some more careful checks then to assert these conditions

pulsipher: Like we should check the variables of disjunct constraints before we add them to ensure they do not contain integer/binary variables

src/reformulate.jl
```diff
@@ -1,82 +1,168 @@
...
-    reformulate_constraint(constr::AbstractArray{<:ConstraintRef}, bin_var, reformulation, param, i, j = missing)
+The `disj` field is the `ConstraintData` object for the disjunction, stored in the 
+`disjunctions` field of the `GDPData` object.
+"""
+# generic fallback (e.g., BigM, Indicator)
+function reformulate_disjunction(model::JuMP.Model, disj::Disjunction, method::AbstractReformulationMethod)
+    ref_cons = Vector{JuMP.AbstractConstraint}()
+    for d in disj.indicators
+        push!(ref_cons, _reformulate_disjunct(model, d, method)...)
+    end
```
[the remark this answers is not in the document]
hdavid16: done

src/reformulate.jl
```diff
@@ -1,82 +1,168 @@
...
+    for d in disj.indicators #reformulate each disjunct
+        _disaggregate_variables(model, d, disj_vrefs, hull) #disaggregate variables for that disjunct
+        push!(ref_cons, _reformulate_disjunct(model, d, hull)...)
+    end
+    for vref in disj_vrefs #create sum constraint for disaggregated variables
+        push!(ref_cons, _aggregate_variable(model, vref, hull))
+    end
 
-    reformulate_constraint(constr::ConstraintRef, bin_var, reformulation, param, i, j = missing, k = missing)
+    return ref_cons
```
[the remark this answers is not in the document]
hdavid16: done

src/reformulate.jl
hdavid16: done

pulsipher:
Note that test should not have its own `Project.toml`. It should use the main one. Any additional dependences can be added with `extra` and `targets` field. See https://github.com/infiniteopt/InfiniteOpt.jl/blob/master/Project.toml

src/hull.jl
hdavid16: Ok. I'll add these checks in the `build_constraint` for the disjunct constraints

hdavid16: done

src/reformulate.jl
```diff
@@ -1,82 +1,163 @@
...
+function reformulate_disjunct_constraint(
+    model::JuMP.Model,  
+    con::Disjunction, 
+    bvref::JuMP.VariableRef,
+    method::AbstractReformulationMethod
+)
+    ref_cons = reformulate_disjunction(model, con, method)
+    new_ref_cons = Vector{JuMP.AbstractConstraint}()
+    for ref_con in ref_cons
+        push!(new_ref_cons, reformulate_disjunct_constraint(model, ref_con, bvref, method)...) #nested = false only for the parent disjunction (only adds the constraints when they have moved up all the way)
```
pulsipher, line 133:
```suggestion
        append!(new_ref_cons, reformulate_disjunct_constraint(model, ref_con, bvref, method)) #nested = false only for the parent disjunction (only adds the constraints when they have moved up all the way)
```

Project.toml
```diff
@@ -1,23 +1,19 @@
...
 
 [deps]
-IntervalArithmetic = "d1acc4aa-44c8-5952-acd4-ba5d80a2a253"
 JuMP = "4076af6c-e467-56ae-b986-b466b2749572"
-Suppressor = "fd094767-a336-5f1f-9728-57cf17d0bbfb"
-Symbolics = "0c5d862f-8b57-4792-8d23-62f2024744c7"
 
 [compat]
-IntervalArithmetic = "0.20"
 JuMP = "1"
```
pulsipher, line 10:
```suggestion
JuMP = "1.15"
```

pulsipher:
I think I figured out how to represent disjunctions at the MOI level, see https://github.com/pulsipher/DisjunctiveProgramming.jl/pull/1

Project.toml
pulsipher: This is needed because JuMP 1.15 introduced `GenericNonlinearExpr`

pulsipher:
I think this is ready to be merged, but it might be worth waiting to see what develops with https://github.com/jump-dev/JuMP.jl/pull/3530 so can we can adjust the logical constraint structure accordingly. Alternatively, we can merge this and then address any adjustments in a future PR before releasing a new version of DisjunctiveProgramming.

hdavid16 (approved):

hdavid16 (approved):

hdavid16:
Copied over to #68

PR #61, opened 2022-12-23
Reference fixes

shivankj11:


paper/ref.bib
```diff
@@ -4,6 +4,7 @@ @inproceedings{agarwal2010automating
   booktitle={International Symposium on Practical Aspects of Declarative Languages},
   pages={134--148},
   year={2010},
+  doi="10.1007/978-3-642-11503-5\_12",
```
hdavid16, line 7: should the backslash \ be there?

hdavid16 (changes requested):
Need to confirm DOI on "Automating mathematical program transformations". I don't think it should have the backslash.

hdavid16 (approved):

paper/ref.bib
hdavid16: Works on overleaf, so leaving as is

PR #71, opened 2023-10-12
Add MOI Disjunction Set Reformulation

pulsipher:
This adds the MOIDisjunction reformulation method which transforms disjunctions in vector constraints that use DisjunctionSet. This enables us to pass disjunctions directly down to the MOI level.

### DIFF
--- a/src/datatypes.jl
+++ b/src/datatypes.jl
@@ -373,6 +373,33 @@ A type for using indicator constraint approach for linear disjunctive constraint
 """
 struct Indicator <: AbstractReformulationMethod end
 
+"""
+    MOIDisjunction <: AbstractReformulationMethod
+A reformulation type for reformulating disjunctions into their MathOptInterface 
+set representations [`DisjucntionSet`](@ref) which are then added to the model.
+"""
+struct MOIDisjunction <: AbstractReformulationMethod end
+
+"""
+    DisjunctionSet{S <: MOI.AbstractSet} <: MOI.AbstractVectorSet
+A MathOptInterface set for representing disjunctions in the format vector of 
+functions in set to enable:
+```julia
+@constraint(model, [funcs...] in DisjunctionSet(n, idxs, sets))
+```
+where the vector of functions `funcs` is a flattened version of all the disjunct 
+constraint functions where the indicator variable is listed first, `n` is the 
+length of `[funcs...]`, `idxs` is a vector of the indices tracking where each 
+disjunct begins (i.e., it stores the indices of the indicator variables in 
+`[funcs...]`), and `sets` is a uses a vector of vectors structure for the MOI sets 
+that correspond to all the disjunct constraint functions.
+"""
+struct DisjunctionSet{S <: _MOI.AbstractSet} <: _MOI.AbstractVectorSet 
+    dimension::Int
+    disjunct_indices::Vector{Int}
+    constraint_sets::Vector{Vector{S}}
+end
+
 ################################################################################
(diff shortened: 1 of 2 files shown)


src/datatypes.jl
```diff
@@ -373,6 +373,33 @@ A type for using indicator constraint approach for linear disjunctive constraint
 """
 struct Indicator <: AbstractReformulationMethod end
 
+"""
+    MOIDisjunction <: AbstractReformulationMethod
+A reformulation type for reformulating disjunctions into their MathOptInterface 
+set representations [`DisjucntionSet`](@ref) which are then added to the model.
```
bernalde, line 379:
```suggestion
set representations [`DisjunctionSet`](@ref) which are then added to the model.
```

pulsipher:
As an update, it seems MOI bridges will likely not work with disjunctions. The problem is that each disjunct can contain an arbitrary number of constraints of any type (including other disjunctions). For a bridge to work, MOI needs to be able to infer want reformulation constraints will be constructed based on the set type. With:
```julia
struct DisjunctionSet{S <: _MOI.AbstractSet} <: _MOI.AbstractVectorSet 
    dimension::Int
    disjunct_indices::Vector{Int}
    constraint_sets::Vector{Vector{S}}
end
```
we will most often end up with something like `DisjuctionSet{MOI.AbstractSet}` which doesn't tell us what type of constraints it contains. Alternatively, we could use a tuple to put all the constraint set types as parametric types:
```julia
struct DisjunctionSet{T <: Tuple} <: _MOI.AbstractVectorSet 
    dimension::Int
    disjunct_indices::Vector{Int}
    constraint_sets::T # flatten everything into a tuple
end
```
Now `T` will contain all constraint set types, but to make a MOI bridge we would have to account for every concrete typed case. This leads to an infinite number of combinations, since there can be any number of disjuncts and constraints with any type. 

So, in the end, it is possible to represent disjunctions in MOI, but bridges won't be practical to implement. However, solvers (e.g., ToQUBO.jl) can still support `DisjunctionSet`s directly. Ultimately, we can make an MOI solver layer similar to ParametricOptInterface that converts `DisjunctionSet`s into forms that other solvers can handle. Moreover, such a layer could support more advanced solution techniques like logic-based outer approximation.

pulsipher:
After talking offline with @hdavid16, another option to is add a canonicalized disjunction set where all the constraints are converted to inequalities first:
```julia
struct DisjunctionSet{T <: Real} <: _MOI.AbstractVectorSet 
    dimension::Int
    disjunct_indices::Vector{Int}
    constraint_sets::Vector{Vector{MOI.LessThan{T}}}
end
```
This would require more processing upfront, change the structure, and potentially exclude some constraint types (e.g., cones), but this would be easy to implement a bridge for.

pulsipher:
@bernalde and @pedromxavier I think this PR gives you the MOI representation that you need. Namely, the addition of `DisjunctionSet <: MOI.AbstractVectorSet`. Let me know your thoughts and if it looks good, I will add tests and we can merge this for you to use.

bernalde:
It looks good to me. I'd be curious to see the tests to see how to interact with the new code.

pedromxavier:
Should there be a `MOI.get` interface for accessing the contents of the Disjunction Set?

pulsipher:
> Should there be a `MOI.get` interface for accessing the contents of the Disjunction Set?

Any suggestions on what the input and output would be precisely?

pulsipher:
Completed via #140

PR #90, opened 2023-10-26
Add `exclusive` option for disjunctions and fix bugs

### LINKED ISSUE #72: Have `@disjunction` automatically add exactly one constraints with a keyword argument
Since 99% of the time this is the case, automating it would be nice. I think we should make it opt-in (i.e., the default would be `false`). 

One other thing to consider is whether the reformulations are still valid if this constraint is not given.

pulsipher:
Closes #72. Some minor bug fixes and changes that came up include:
- Have deletion methods for disjunctions and disjunct constraints remove all their mappings
- Remove unnecessary `GDPData(args...)` method
- Include the macro docstrings in the docs
- Moving reformulation specific methods
- Fixing the reformulation methods for exactly constraints to better handle assumptions on function type (more work is needed)
- Improved error messaging for incorrect keyword arguments to `@disjunction`
- Fixed some test functions to use `@test` on the first call of a method/macro to be tested which leads to more helpful error messages when it doesn't work (more work is needed to do this with all tests)


hdavid16:
I think we should use a different name for this keyword argument since it is not exactly 1 for Nested disjunctions.

pulsipher:
> I think we should use a different name for this keyword argument since it is not exactly 1 for Nested disjunctions.

Any suggestions?

hdavid16 (approved):

PR #92, opened 2023-10-30
Disallow logical propositions with only 1 logical variable

hdavid16:
Currently, `@constraint(m, y := true)` throws an error, but `@constraint(m, logical_not(y) := true` does not. For consistency, both should error since the user should `fix` the value of the logical variable instead. This PR makes adds additional checks to enforce this.

### DIFF
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -493,14 +493,23 @@ end
 
 # Check that logical expression is valid
 function _check_logical_expression(ex)
+    _check_logical_expression_literal(ex)
+    _check_logical_expression_operator(ex)
+end
+function _check_logical_expression_literal(ex::_LogicalExpr)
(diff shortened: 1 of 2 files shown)


pulsipher (changes requested):
This is good. The only thing is that with these changes I think the following is unnecessary: https://github.com/hdavid16/DisjunctiveProgramming.jl/blob/b26e1ed9046feef832f5b0ac7bf50138c53b7144/src/logic.jl#L275-L276. This is the line that is missing the code coverage.

hdavid16:
It is still needed to reformulate:
```julia
@constraint(model, y[1] ∧ ¬y[2] := true)
```
I have updated `test_intersection_reformulation()` to use this constraint for the test, which should allow for testing the line you indicate.

PR #96, opened 2023-12-05
Support `JuMP.value` for Logical Variables

### LINKED ISSUE #94: Support `JuMP.value` for `LogicalVariableRef`
We should add `JuMP.value(vref::LogicalVariableRef) = JuMP.value(binary_variable(vref))` to query the values as expected. 

On a similar note, we may want to consider supporting various constraint methods for disjunct constraints.

pulsipher:
Closes #94. It also alphabetizes the Project file to fix an Aqua test failure.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -8,10 +8,10 @@ JuMP = "4076af6c-e467-56ae-b986-b466b2749572"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 
 [compat]
+Aqua = "0.8"
 JuMP = "1.16"
 Reexport = "1"
 julia = "1.6"
-Aqua = "0.8"
 
 [extras]
 Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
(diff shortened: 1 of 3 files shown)


hdavid16:
Thanks for putting this together. What are your thoughts on returning the corresponding Bool instead of 0 or 1? Should be a simple update: value(binary) == 1 (or use isone).

pulsipher:
> Thanks for putting this together. What are your thoughts on returning the corresponding Bool instead of 0 or 1? Should be a simple update: value(binary) == 1 (or use isone).

I thought about this, but the problem is MIP solvers return binary variable solutions as `Float64` not 0 or 1. So, we would need to use some sort of tolerance to round to the appropriate integer solution and then convert to `Bool`. I would prefer to not hard-code some tolerance or add more complexity to `value` by adding a keyword that would only work on logical variables.

hdavid16:
That is a good point. Since the value is binary, what about comparing it to 0.5 (value > 0.5)?

pulsipher:
Ok, I have changed it to round based on `0.5`.

hdavid16 (approved):

PR #109, opened 2024-01-18
Support disjunct constraints with integer variables

### LINKED ISSUE #105: Disjunct constraints cannot contain binary or integer variables.
I am surprised that I run into this error when I define my times as `days + timeofday` where `days` are Integers and `timeofday` is Float.

I want to use DisjunctiveProgramming.jl to define XOR constraints on time intervals.

Is there a theoretical limitation why such composed expressions with Integer/Binary component are not compatible with DisjunctiveProgramming.jl?

hdavid16:
Fixes #105. Still disallows binary variables in a disjunct constraint. The error message is improved to indicate that the user should rely on Logic variables to model any binary decisions.

### DIFF
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -26,8 +26,8 @@ _vec_to_scalar_set(::_MOIAtMost) = _MOI.LessThan
 ################################################################################
 #                         BOILERPLATE EXTENSION METHODS
 ################################################################################
-for (RefType, loc) in ((:DisjunctConstraintRef, :disjunct_constraints), 
-                       (:DisjunctionRef, :disjunctions), 
+for (RefType, loc) in ((:DisjunctConstraintRef, :disjunct_constraints),
+                       (:DisjunctionRef, :disjunctions),
                        (:LogicalConstraintRef, :logical_constraints))
     @eval begin
         @doc """
@@ -81,7 +81,7 @@ for (RefType, loc) in ((:DisjunctConstraintRef, :disjunct_constraints),
         @doc """
             JuMP.constraint_object(cref::$($RefType))
 
-        Return the underlying constraint data for the constraint 
+        Return the underlying constraint data for the constraint
             referenced by `cref`.
         """
         function JuMP.constraint_object(cref::$RefType)
@@ -123,7 +123,7 @@ function JuMP.delete(model::JuMP.AbstractModel, cref::DisjunctionRef)
         delete!(exactly1_dict, cref)
     end
     _set_ready_to_optimize(model, false)
-    return 
+    return
 end
 
 """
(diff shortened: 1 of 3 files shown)


src/constraints.jl
```diff
@@ -150,25 +150,28 @@ function JuMP.delete(model::JuMP.AbstractModel, cref::LogicalConstraintRef)
...
     vars = Set{JuMP.variable_ref_type(expr)}()
-    _interrogate_variables(v -> push!(vars, v), expr) 
-    if any(JuMP.is_binary.(vars)) || any(JuMP.is_integer.(vars))
-        error("Disjunct constraints cannot contain binary or integer variables.")
+    _interrogate_variables(v -> push!(vars, v), expr)
+    if any(JuMP.is_binary.(vars))
+        error(
+            "Disjunct constraints should not contain binary variables. " *
+            "Instead, use Logical variables to express any binary decisions."
+        )
```
pulsipher, line 166: I am not sure this is sufficient since logical variables cannot be used inside algebraic constraints with regular JuMP variables. I think allowing both binary and integer variables is the way to go. Naturally, this has implications for the reformulations where we currently assume any binary variable is a disjunct indicator variable. However, we can instead check if the binary variable corresponds to an indicator variable via a mapping in `GDPData`.

pulsipher:
The printing problem can be fixed by updating to use the printing API provided by https://github.com/jump-dev/JuMP.jl/pull/3609. I will open a PR when I can find some extra time.

src/constraints.jl
hdavid16: My point here is that if the user wants to have a binary variable inside a disjunct constraint, this should be coming from a nested disjunction. The user should model their problem as a proper GDP. Perhaps a better error message would be:

"Disjunct constraints can only contain continuous or integer variables, but not binary variables. If you are modeling binary decisions, you should define these in a nested disjunction."

I am still not convinced that we should allow binary variables in a disjunction constraint. I see integer variables being allowed, as these can have a physical meaning (e.g., number of trucks purchased), but not binary variables.

hdavid16: @pulsipher I have updated the error message text to:
```julia
        error(
            "Disjunct constraints can only contain continuous or integer variables, " *
            "not binary variables. If you are modeling binary decisions, " *
            "you should express these using the Logical variables that serve as indicators " *
            "to each of the disjuncts. If these are lower-level decisions to the disjunct " *
            "being defined, these can be modeled using nested disjunctions."
        )
```

What are your thoughts on this issue of allowing binaries?

pulsipher: I still suspect that are situations where it is difficult to reformulate into nested disjunctions. For instance, certain conic sets that rely on binary variables or perhaps a piecewise linear function. These could be reformulated into nested disjunctions, but that puts a lot of burden on the user. 

However, for now adding integer variables is still a win and the error message makes things more clear. We can wait for someone to come up with a compelling use case for allowing binary variables.

pulsipher (approved):
I think we're ok to merge this.

PR #118, opened 2024-10-01
Support Logical Complements

### LINKED ISSUE #106: Is it possible to save a Logical variable if I have a disjunction of merely two alternatives?
Hi there,

the documentation advises to create on logical variable per `Disjunct`.
I am wondering, whether this can be simplified in the case of only two `Disjunct`s.

Maybe one can reduce the number of needed variables down from 2 to 1 because one Logical variable might represent two options.

pulsipher:
This PR closes #106. It does so by adding the `logical_complement` keyword argument when creating logical variables. For instance, we can have:
```julia
@variable(model, Y1, Logical) # binary_variable(Y1) returns a single binary
@variable(model, Y2, Logical, logical_complement = Y1) # binary_variable(Y2) = 1 - binary_variable(Y1)
```
These logical variables can then be used as normal to build disjunctions, but the underlying binary expressions avoid adding an unnecessary variable.

This works with all reformulations, but currently has the limitation that logical variables used inside nested disjunctions cannot be logical complements.

### DIFF
--- a/src/reformulate.jl
+++ b/src/reformulate.jl
@@ -78,8 +78,9 @@ function _reformulate_disjunctions(model::JuMP.AbstractModel, method::AbstractRe
         dref = DisjunctionRef(model, idx)
         if requires_exactly1(method) && !haskey(gdp_data(model).exactly1_constraints, dref) &&
             !any(has_logical_compliment.(disj.constraint.indicators))
-            error("Reformulation method `$method` requires disjunctions where only 1 disjunct is selected, " *
-                  "but `exactly1 = false` for disjunction `$dref`.")
+            error("Reformulation method `$method` requires disjunctions where only 1 disjunct is selected, ",
+                  "but `exactly1 = false` or none of the logical variables are logical compliments ",
+                  "for disjunction `$dref`.")
         end
         if requires_variable_bound_info(method)
             for vref in _get_disjunction_variables(model, disj.constraint)
--- a/src/reformulate.jl
+++ b/src/reformulate.jl
@@ -76,7 +76,8 @@ function _reformulate_disjunctions(model::JuMP.AbstractModel, method::AbstractRe
     for (idx, disj) in _disjunctions(model)
         disj.constraint.nested && continue #only reformulate top level disjunctions
         dref = DisjunctionRef(model, idx)
-        if requires_exactly1(method) && !haskey(gdp_data(model).exactly1_constraints, dref)
+        if requires_exactly1(method) && !haskey(gdp_data(model).exactly1_constraints, dref) &&
+            !any(has_logical_compliment.(disj.constraint.indicators))
             error("Reformulation method `$method` requires disjunctions where only 1 disjunct is selected, " *
                   "but `exactly1 = false` for disjunction `$dref`.")
         end
@@ -94,7 +95,7 @@ function _reformulate_disjunctions(model::JuMP.AbstractModel, method::AbstractRe
     end
 end
 
-# disjuncts
+# individual disjunctions
 """
     reformulate_disjunction(
         model::JuMP.AbstractModel, 
@@ -137,7 +138,7 @@ end
     reformulate_disjunct_constraint(
         model::JuMP.AbstractModel,  
         con::JuMP.AbstractConstraint, 
-        bvref::JuMP.AbstractVariableRef,
+        bvref::Union{JuMP.AbstractVariableRef, JuMP.GenericAffExpr},
         method::AbstractReformulationMethod
     )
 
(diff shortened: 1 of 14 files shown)


pulsipher:
Later on, we could add the syntax:
```julia
@variable(model, Y1, Logical)
@variable(model, Y2 == ¬Y1, Logical) 
```
to make it more symbolic, but that can wait for another PR.

src/reformulate.jl
```diff
@@ -76,7 +76,8 @@ function _reformulate_disjunctions(model::JuMP.AbstractModel, method::AbstractRe
     for (idx, disj) in _disjunctions(model)
         disj.constraint.nested && continue #only reformulate top level disjunctions
         dref = DisjunctionRef(model, idx)
-        if requires_exactly1(method) && !haskey(gdp_data(model).exactly1_constraints, dref)
+        if requires_exactly1(method) && !haskey(gdp_data(model).exactly1_constraints, dref) &&
+            !any(has_logical_compliment.(disj.constraint.indicators))
             error("Reformulation method `$method` requires disjunctions where only 1 disjunct is selected, " *
```
hdavid16, line 81: This error should be updated to note that `exactly1 = false` or no logical compliments are defined for disjunction...

src/variables.jl
```diff
@@ -27,12 +34,18 @@ function JuMP.build_variable(
         _error("Invalid fix value, must be false or true.")
     elseif info.has_start && !isone(info.start) && !iszero(info.start)
         _error("Invalid start value, must be false or true.")
+    elseif (info.has_start || info.has_fix) && !isnothing(logical_compliment)
+        _error("Cannot fix or provide a start value for logical variable " *
```
hdavid16, line 38: "for a logical variable that is a logical compliment variable" (added "a" and "is")

src/variables.jl
```diff
@@ -27,12 +34,18 @@ function JuMP.build_variable(
         _error("Invalid fix value, must be false or true.")
     elseif info.has_start && !isone(info.start) && !iszero(info.start)
         _error("Invalid start value, must be false or true.")
+    elseif (info.has_start || info.has_fix) && !isnothing(logical_compliment)
```
hdavid16, line 37: Why? Is it because we haven't added this support yet or because it cannot be done?

src/variables.jl
```diff
@@ -27,12 +34,18 @@ function JuMP.build_variable(
...
+        _error("Cannot fix or provide a start value for logical variable " *
+               "that has a logical compliment variable.")
+    elseif !isnothing(logical_compliment) && has_logical_compliment(logical_compliment)
+        _error("Cannot specify a logical compliment that itself is a " *
```
hdavid16, line 41: Maybe add something about using the logical_complement flag on only one of the variables in the disjunction

src/variables.jl
```diff
@@ -203,7 +226,10 @@ function JuMP.set_start_value(
     vref::LogicalVariableRef, 
     value::Union{Nothing, Bool}
     )
-    new_var = LogicalVariable(JuMP.fix_value(vref), value)
+    if has_logical_compliment(vref)
+        error("Cannot set the start value of a variable with a logical compliment.")
```
hdavid16, line 230: yet?

src/variables.jl
```diff
@@ -236,7 +262,10 @@ constraint if one exists, otherwise create a
 new one.
 """
 function JuMP.fix(vref::LogicalVariableRef, value::Bool)
-    new_var = LogicalVariable(value, JuMP.start_value(vref))
+    if has_logical_compliment(vref)
+        error("Cannot fix value of a variable with a logical compliment.")
```
hdavid16, line 266: yet?

hdavid16 (changes requested):
Looks great. Thanks for putting this together. Added some minor comments. We can merge after they are addressed.

src/variables.jl
pulsipher: Because a logical compliment variable Y2 actually maps to 1 - Y1. Any start value or fix value with Y2 would conflict any defined in Y1. We could add a bunch of checks to avoid such conflicts, but that seems unnecessary. I clarified the error message to make this more clear.

src/variables.jl
pulsipher: This is to avoid conflicts. I clarified the message.

src/variables.jl
pulsipher: This is to avoid conflicts. I clarified the error.

pulsipher:
Thanks, I have addressed the comments. The restriction on not specifying the fix or start value is to avoid conflicts with the variable it is the compliment of.

pulsipher:
> I think it would be better to say "Fix its logical complement instead". Using "Try" makes it sound like that might not work. Also complement should be spelled with an `e` everywhere (not an `i`...we don't like compliments XD)

Done.

hdavid16 (approved):
Great. Thanks!

PR #120, opened 2025-06-25
Support Multiple Big-M reformulation.

dnguyen227:
This PR adds the Multiple Big-M reformation technique. The technique can be called similar to BigM and Hull:

`optimize!(model, gdp_method = MBM(Gurobi.Optimizer))`

In `src/mbm.jl`  are extensions for `reformulate_disjunction`, `_reformulate_disjunct`, and `reformulate_disjunct_constraint` functions that dispatch over the new MBM datatype.

Documentation to follow in another PR.


src/mbm.jl
```diff
@@ -0,0 +1,329 @@
...
+
+#Solve a mini-model to find the maximum value of the objective function for M value
+function _mini_model(
+    model::JuMP.AbstractModel, 
+    objective::ScalarConstraint{T,S}, 
+    constraints::Vector{DisjunctConstraintRef}, 
+    method::MBM
+) where {T,S <: Union{_MOI.LessThan, _MOI.GreaterThan}}
+    sub_model = Model()
+    new_vars = Dict{VariableRef, VariableRef}()
```
pulsipher, line 225: This will be potentially problematic if we try to use this with the InfiniteOpt extension. Since it doesn't use `JuMP.Model` or `JuMP.VariableRef`. Also stylistically, please preface all JuMP methods/types with `JuMP.`.

src/mbm.jl
```diff
@@ -0,0 +1,329 @@
...
+function replace_variables_in_constraint(fun::Number, var_map::Dict{VariableRef, VariableRef})
+    return fun
+end
+
+function replace_variables_in_constraint(fun::NonlinearExpr, var_map::Dict{VariableRef, VariableRef})
+    new_args = Any[replace_variables_in_constraint(arg, var_map) for arg in fun.args]
+    return JuMP.NonlinearExpr(fun.head, new_args)
+end
+
+function replace_variables_in_constraint(fun::Vector{T}, var_map::Dict{VariableRef, VariableRef}) where T
```
pulsipher, line 323: See above comment about `VariableRef`.

pulsipher (changes requested):
Overall pretty good, please generalize to work with the InfiniteOpt extension and add this to the docs.

src/constraints.jl
```diff
@@ -23,6 +23,32 @@ _vec_to_scalar_set(::_MOIExactly) = _MOI.EqualTo
...
+    JuMP.is_integer(original_var) && JuMP.set_integer(new_var)
+    JuMP.is_binary(original_var) && JuMP.set_binary(new_var)
+    
+    # Handle fixed values with force=true (as in original MBM code)
+    if JuMP.is_fixed(original_var)
+        JuMP.fix(new_var, JuMP.fix_value(original_var); force=true)
+    end
+    
+    return new_var
+end
```
dnguyen227, line 48: This new function differs from` make_disaggregated_variable()`. The main difference is that instead of passing the variable attributes, I instead pass the variable itself and check for the attributes within this new function.

Why?
The alternative to this would be to have all attributes as possible inputs and default them to `nothing`, but then the function would still need to check if the value is 'nothing'.
For example:
```
function _copy_variable(
    target_model::JuMP.AbstractModel,
    original_var::JuMP.AbstractVariableRef;
    base_name::Union{String, Nothing} = nothing,
    lower_bound::Union{Float64, Nothing} = nothing,
    upper_bound::Union{Float64, Nothing} = nothing,
    start_value::Union{Float64, Nothing} = nothing,
    is_integer::Union{Bool, Nothing} = nothing,
    is_binary::Union{Bool, Nothing} = nothing,
    fixed_value::Union{Float64, Nothing} = nothing
    )
    
    # Create new variable with base name
    var_name = base_name !== nothing ? base_name : JuMP.name(original_var)
    new_var = JuMP.@variable(target_model, base_name = var_name)

    # Set attributes if provided
    lower_bound !== nothing && JuMP.set_lower_bound(new_var, lower_bound)
    upper_bound !== nothing && JuMP.set_upper_bound(new_var, upper_bound)
    start_value !== nothing && JuMP.set_start_value(new_var, start_value)
    is_integer !== nothing && is_integer && JuMP.set_integer(new_var)
    is_binary !== nothing && is_binary && JuMP.set_binary(new_var)
    
    # Handle fixed value with force=true
    if fixed_value !== nothing
        JuMP.fix(new_var, fixed_value; force=true)
    end

    return new_var
end
```

This involved two checks, once before passing to the function, then one inside the function. So I thought just passing the variable to limit it to one check was better.

Let me know if there is a better way to do this, especially if I can make more akin to `make_disaggregated_variable`.

src/mbm.jl
```diff
@@ -0,0 +1,317 @@
...
+end
+
+#Solve a mini-model to find the maximum value of the objective function for M value
+function _mini_model(
+    model::JuMP.AbstractModel, 
+    objective::JuMP.ScalarConstraint{T,S}, 
+    constraints::Vector{DisjunctConstraintRef}, 
+    method::MBM
+) where {T,S <: Union{_MOI.LessThan, _MOI.GreaterThan}}
+    sub_model = JuMP.Model()
```
dnguyen227, line 227: Needs a helper method to create different types of models instead of just a generic JuMP model.

pulsipher:
Let's wait for #123 to be merged before finalizing this.

src/constraints.jl
```diff
@@ -23,6 +23,34 @@ _vec_to_scalar_set(::_MOIExactly) = _MOI.EqualTo
...
+    JuMP.is_integer(original_var) && JuMP.set_integer(new_var)
+    JuMP.is_binary(original_var) && JuMP.set_binary(new_var)
+    
+    if JuMP.is_fixed(original_var)
+        JuMP.fix(new_var, JuMP.fix_value(original_var); force=true)
+    end
+    
+    return new_var
+end
+
```
pulsipher, line 46: Shouldn't this be removed now?

src/mbm.jl
```diff
@@ -0,0 +1,405 @@
...
+    ref_cons::Vector{JuMP.AbstractConstraint}, 
+    lvref::LogicalVariableRef,
+    conlvref::Vector{LogicalVariableRef},
+    method::AbstractReformulationMethod
+)
+    
+    bconref = Dict(d => binary_variable(d) for d in conlvref)
+    !haskey(_indicator_to_constraints(model), lvref) && return
+    constraints = _indicator_to_constraints(model)[lvref]
+    M = Dict{LogicalVariableRef,Float64}()
```
pulsipher, line 50: Instead of `Float64`, we should use the `value_type`. Please correct all the instances of this.

pulsipher (changes requested):
Almost there. Also, please verify over some example problems that this returns the correct answer (e.g., the same as big-M).

src/mbm.jl
```diff
@@ -0,0 +1,454 @@
...
+            if con isa JuMP.ScalarConstraint{<:Any, <:MOI.Interval}
+                # Create lower and upper bound constraints
+                lower_con = JuMP.build_constraint(error, con.func, 
+                    MOI.GreaterThan(con.set.lower))
+                upper_con = JuMP.build_constraint(error, con.func, 
+                    MOI.LessThan(con.set.upper))
+                # Create new disjunct constraints
+                JuMP.add_constraint(model,_DisjunctConstraint(lower_con, d))
+                JuMP.add_constraint(model,_DisjunctConstraint(upper_con, d))
+                JuMP.delete(model, cref)
```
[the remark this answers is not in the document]
dnguyen227: I am not sure why my past self has chosen to do this. I tried approach to EqualTo just now and it works fine. Am going to run a few extra problems to check the reformulation as a sanity check but this looks like an easy fix.

pulsipher (changes requested):
This looks good, let's address the todo in the tests and we are good.

PR #123, opened 2025-09-18
Add Extension Friendly Variable Creation Methods

dnguyen227:
For the purposes of copying variables from one model to another, while also having the flexibility to modify properties before copy with the VariableProperties. Simple example below.

```
using JuMP
import DisjunctiveProgramming as DP

# Create original variable with properties
original_model = Model()
@variable(original_model, 2 <= x <= 10, start = 5.0)

# Extract variable properties
props = DP.VariableProperties(x)

#Capability to change property before  copying onto another model.
props.is_integer = true

# Create new model and recreate variable with same properties
new_model = Model()
x_copy = DP._variable_from_properties(new_model, props)

```

### DIFF
--- a/src/hull.jl
+++ b/src/hull.jl
@@ -1,15 +1,6 @@
 ################################################################################
 #                              VARIABLE DISAGGREGATION
 ################################################################################
-"""
-    requires_disaggregation(vref::JuMP.AbstractVariableRef)::Bool
-
-Return a `Bool` whether `vref` requires disaggregation for the [`Hull`](@ref) 
-reformulation. This is intended as an extension point for interfaces with 
-DisjunctiveProgramming that use variable reference types that are not 
-`JuMP.GenericVariableRef`s. Errors if `vref` is not a `JuMP.GenericVariableRef`.
-See also [`make_disaggregated_variable`](@ref).
-"""
 requires_disaggregation(vref::JuMP.GenericVariableRef) = true
 function requires_disaggregation(::V) where {V}
     error("`Hull` method does not support expressions with variable " *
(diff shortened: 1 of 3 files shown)


pulsipher:
This is a good start, I would suggest the following:
```julia
# Stores all the info that could be used to create a JuMP variable
mutable struct VariableProperties{L, U, F, S, SET, T}
    info::JuMP.VariableInfo{L, U, F, S}
    name::String
    set::SET
    variable_type::T # this is critical for extensions
end 

# Default JuMP constructor (extensions will need to define this)
function VariableProperties(vref::JuMP.GenericVariableRef{T}) where T
    info = JuMP.VariableInfo(
        JuMP.has_lower_bound(vref),
        JuMP.has_lower_bound(vref) ? JuMP.lower_bound(vref) : zero(T),
        JuMP.has_upper_bound(vref),
        JuMP.has_upper_bound(vref) ? JuMP.upper_bound(vref) : zero(T),
        JuMP.is_fixed(vref),
        JuMP.is_fixed(vref) ? JuMP.fix_value(vref) : zero(T),
        !isnothing(JuMP.start_value(vref)),
        JuMP.start_value(vref),
        JuMP.is_binary(vref),
        JuMP.is_integer(vref)
    )
    name = JuMP.name(vref)
    set = JuMP.is_variable_in_set(vref) ? JuMP.moi_set(JuMP.constraint_object(JuMP.VariableInSetRef(vref))) : nothing
    return VariableProperties(info, name, set, nothing)
end

# Function for converting VariableProperties into variables (can be extended, but set up to avoid needing to do so)
function create_variable(model::JuMP.AbstractModel, props::VariableProperties)
    var = _make_variable_object(props)
    if !isnothing(props.set)
        var = JuMP.build_variable(error, var, props.set)
    end
    return JuMP.add_variable(model, var, props.name)
end

function _make_variable_object(props::VariableProperties{L, U, F, S, SET, Nothing}) where {L, U, F, S, SET}
    return JuMP.build_variable(error, props.info)
end
function _make_variable_object(props::VariableProperties)
    return JuMP.build_variable(error, props.info, props.variable_type)
end
```

Then we can use this to define useful functions for the reformulations, like copy:
```julia
function variable_copy(model::JuMP.AbstractModel, vref::JuMP.AbstractVariableRef)
    props = VariableProperties(vref)
    return create_variable(model, props)
end 
```

Such that we can do something like:
```julia
using JuMP

m = Model()
@variable(m, 0 <= x <= 4, Int)
@variable(m, p in Parameter(42))
x_copy = copy_variable(m, x)
m2 =Model()
p_copy = copy_variable(m2, p)
```

pulsipher (changes requested):
This is coming along. This PR should also update the existing transformation methods (e.g., Hull) to use these methods as appropriate.

pulsipher (changes requested):
Almost there, just a few minor things.

dnguyen227:
> Almost there, just a few minor things and need to fix the tests.

Please give the workflows another try 🤞

PR #124, opened 2025-10-03
Support P-Split Reformulation for (affine and quadratic constraints only)

dnguyen227:
This PR adds the P-Split reformulation technique. The technique is called with a vector of partitions:
For example for two partitions (p=2):

model = GDPModel()
...
@variable(model, x[1:4], Disjunct(...))
...
partition = [ [x[1] , x[2]], [x[3] , x[4]]] #one partition contains x1,x2 and the other contains x3,x4
optimize!(model, gdp_method = PSplit(partition))

In src/psplit.jl are extensions for reformulate_disjunct_constraint as well as new functions for building partitioned expressions.


pulsipher:
@dnguyen227, is this ready for review?

dnguyen227:
Yes please review.

src/psplit.jl
```diff
@@ -0,0 +1,430 @@
...
+################################################################################
+#                              BOUND AUXILIARY
+################################################################################
+function _bound_auxiliary(
+    model::JuMP.AbstractModel,
+    v::JuMP.AbstractVariableRef,
+    func::JuMP.GenericAffExpr,
+    method::PSplit
+)   
+    T = JuMP.value_type(typeof(model))
```
pulsipher, line 63: You can get `T` by using `func::JuMP.GenericAffExpr{T, V}`

src/psplit.jl
```diff
@@ -0,0 +1,430 @@
...
+    JuMP.set_upper_bound(v, upper_bound)
+end    
+
+function _bound_auxiliary(
+    model::JuMP.AbstractModel,
+    v::JuMP.AbstractVariableRef,
+    func::Number,
+    method::PSplit
+) 
+    #Do nothing?
```
pulsipher, line 89: just put an empty `return`

src/psplit.jl
```diff
@@ -0,0 +1,430 @@
...
+end
+
+function _bound_auxiliary(
+    model::JuMP.AbstractModel,
+    v::JuMP.AbstractVariableRef,
+    func::JuMP.GenericQuadExpr,
+    method::PSplit
+) 
+    T = JuMP.value_type(typeof(model))
```
pulsipher, line 98: Again you can get `T` from `func`

src/psplit.jl
```diff
@@ -0,0 +1,430 @@
...
+            var_lb, var_ub = variable_bound_info(var)
+            if coeff > 0.0
+                lower_bound += coeff * var_lb
+                upper_bound += coeff * var_ub
+            else
+                lower_bound += coeff * var_ub
+                upper_bound += coeff * var_lb
+            end
+        end
+    end
```
pulsipher, line 115: Is there a way to refactor so this isn't repeated code from the affine function above?

src/psplit.jl
```diff
@@ -0,0 +1,430 @@
...
+    JuMP.set_upper_bound(v, upper_bound)
+end
+
+function _bound_auxiliary(
+    model::JuMP.AbstractModel,
+    v::JuMP.AbstractVariableRef,
+    func::JuMP.AbstractVariableRef,
+    method::PSplit
+) 
+    T = JuMP.value_type(typeof(model))
```
pulsipher, line 153: you can get `typeof(model)` from `model::M`

pulsipher (changes requested):
I left some initial comments for improvement. One fundamental observation though is that P-Split just applies Hull to each partitioned subset. With this in mind Pyomo simply applies their Hull transformation after getting the partitions: https://github.com/Pyomo/pyomo/blob/main/pyomo/gdp/plugins/between_steps.py. I think we should do same, that way we aren't unnecessarily repeating the Hull code.  Moreover, the existing Hull codebase is already battletested.

src/datatypes.jl
```diff
@@ -431,6 +431,65 @@ mutable struct _Hull{V <: JuMP.AbstractVariableRef, T} <: AbstractReformulationM
...
+    
+    return PSplit(partition)
+    end
+end
+
+# temp struct to store variable disaggregations (reset for each disjunction)
+mutable struct _PSplit{V <: JuMP.AbstractVariableRef, M <: JuMP.AbstractModel} <: AbstractReformulationMethod
+    partition::Vector{Vector{V}}
+    sum_constraints::Dict{LogicalVariableRef{M}, Vector{<:AbstractConstraint}}
+    hull::_Hull
```
pulsipher: _Hull needs to be fully typed with its parameteric types

src/hull.jl
```diff
@@ -64,14 +65,23 @@ end
...
+    # push!(ref_cons, JuMP.build_constraint(error, con_expr, _MOI.EqualTo(0)))
+
+    expr = Base.zero(JuMP.GenericAffExpr{JuMP.value_type(M), V})
+    JuMP.add_to_expression!(expr, -1.0, vref)
+    #TODO: One(T)
+    for dv in method.disjunction_variables[vref]
+        JuMP.add_to_expression!(expr, 1.0, dv)
+    end
+    
+    push!(ref_cons, JuMP.build_constraint(error, expr, _MOI.EqualTo(0)))
```
pulsipher: Why is this change needed?

pulsipher (approved):

PR #125, opened 2025-10-28
Support cutting planes technique.

dnguyen227:
This PR adds the cutting planes reformulation technique. The method iteratively generates cutting planes using a separation problem and relaxed Big-M formulation before applying a final reformulation.

```julia
model = GDPModel()
@variable(model, x[1:2])
@variable(model, Y[1:2], Logical)
@constraint(model, x[1] <= 3, Disjunct(Y[1]))
@constraint(model, x[1] <= 4, Disjunct(Y[2]))
@disjunction(model, [Y[1], Y[2]])
@objective(model, Max, sum(x))

# Use with default settings (3 iterations, BigM final reformulation)
optimize!(model, gdp_method = cutting_planes(HiGHS.Optimizer))

# Or customize parameters
optimize!(model, gdp_method = cutting_planes(HiGHS.Optimizer;max_iter=10, seperation_tolerance=1e-4, final_reform_method=Indicator(), M_value=1e6))
```

### DIFF
--- a/test/constraints/cuttingplanes.jl
+++ b/test/constraints/cuttingplanes.jl
@@ -54,17 +54,13 @@ function test_solve_SEP()
     main_to_rBM_map = Dict(v => rBM_ref_map[v] for v in all_variables(model))
     JuMP.set_optimizer(SEP, method.optimizer)
     JuMP.set_optimizer(rBM, method.optimizer)
+    JuMP.set_silent(rBM)
+    JuMP.set_silent(SEP)
+    JuMP.relax_integrality(rBM)
+    JuMP.relax_integrality(SEP)
     JuMP.@objective(rBM, sense, 
     DP._replace_variables_in_constraint(obj, main_to_rBM_map)
     )
-    for m in [SEP, rBM]
-        binary_vars = filter(is_binary, all_variables(m))
-        for var in binary_vars
-            unset_binary(var)
-            set_lower_bound(var, 0.0)
-            set_upper_bound(var, 1.0)
-        end
-    end
     rBM_to_SEP_map = Dict{var_type, var_type}()
     SEP_to_rBM_map = Dict{var_type, var_type}()
     for (var, rBM_var) in main_to_rBM_map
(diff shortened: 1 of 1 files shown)


src/cuttingplanes.jl
```diff
@@ -0,0 +1,138 @@
...
+    _replace_variables_in_constraint(obj, main_to_rBM_map)
+    )
+    for m in [SEP, rBM]
+        binary_vars = filter(is_binary, all_variables(m))
+        for var in binary_vars
+            unset_binary(var)
+            set_lower_bound(var, 0.0)
+            set_upper_bound(var, 1.0)
+        end
+    end
```
pulsipher: Instead of this, you can simply use `JuMP.relax_integrality`

src/cuttingplanes.jl
```diff
@@ -0,0 +1,138 @@
...
+    rBM_to_SEP_map = Dict{var_type, var_type}()
+    SEP_to_rBM_map = Dict{var_type, var_type}()
+    for (var, rBM_var) in main_to_rBM_map
+        SEP_var = main_to_SEP_map[var]
+        rBM_to_SEP_map[rBM_var] = SEP_var
+        SEP_to_rBM_map[SEP_var] = rBM_var
+    end
+    
+    i = 1
+    sep_obj = 1
```
pulsipher: We should probably initialize this to `Inf`

src/cuttingplanes.jl
```diff
@@ -0,0 +1,138 @@
...
+    end
+    reformulate_model(model, method.final_reform_method)
+    return
+end
+
+function _solve_rBM(
+    rBM::M,
+) where {M <: JuMP.AbstractModel}
+    T = JuMP.value_type(M)
+    JuMP.set_silent(rBM)
```
pulsipher: Do we need to set this on each solve?

src/cuttingplanes.jl
```diff
@@ -0,0 +1,138 @@
...
+    SEP_sol::Dict{<:JuMP.AbstractVariableRef, T},
+) where {M <: JuMP.AbstractModel, T <: Number}
+    main_vars = JuMP.all_variables(model)
+
+    ξ_sep = Dict{JuMP.AbstractVariableRef, T}(var => zero(T) for var in main_vars)
+    for var in main_vars
+        ξ_sep[var] = 2*(SEP_sol[main_to_SEP_map[var]] - rBM_sol[main_to_rBM_map[var]])
+    end
+    main_cut = JuMP.@expression(model, 
+    sum(ξ_sep[var]*(var - SEP_sol[main_to_SEP_map[var]]) for var in main_vars)
```
pulsipher: this should be indented

PR #126, opened 2025-10-31
Addition of copy_model and copy_gdp_data to copy GDP models.

dnguyen227:
This PR extends `JuMP.copy_extension_data` in order to allow `GDPModel`s to be an input to `JuMP.copy_model()`. 
`JuMP.copy_extension_data` does not receive a reference map as one of it's argument so it's impossible to recover disjunct constraints. Therefore another function was made to be called (`DP.copy_gdp_data()`) in order to populate the new model's `GDPData`. The workflow goes sequentiallty as in the example below.

```julia
using HiGHS, JuMP
import DisjunctiveProgramming as DP  
m = DP.GDPModel(HiGHS.Optimizer)
@variable(m, 0 ≤ x[1:2] ≤ 20)
@variable(m, Y[1:2], DP.Logical)
@constraint(m, [i = 1:2], [2,5][i] ≤ x[i] ≤ [6,9][i], DP.Disjunct(Y[1]))
@constraint(m, [i = 1:2], [8,10][i] ≤ x[i] ≤ [11,15][i], DP.Disjunct(Y[2]))
DP.@disjunction(m, Y)
@objective(m, Max, sum(x))

m_copy, ref_map = JuMP.copy_model(m)
lv_map = DP.copy_gdp_data(m, m_copy, ref_map)
set_optimizer(m, HiGHS.Optimizer)
set_optimizer_attribute(m, "output_flag", false) 

set_optimizer(m_copy, HiGHS.Optimizer)
set_optimizer_attribute(m_copy, "output_flag", false)

optimize!(m, gdp_method = DP.BigM())
optimize!(m_copy, gdp_method = DP.Hull())
```

In addition there is also `copy_model_and_gdp_data` to do both
An alternative would be to extend JuMP.copy_model itself but at a glance I believe there might be consequences when we want to have DP coexist with other packages and have the ability to copy.

### DIFF
--- a/src/utilities.jl
+++ b/src/utilities.jl
@@ -8,6 +8,18 @@ function _copy_model(
     return M()
 end
 
+"""
+    JuMP.copy_extension_data(
+        data::GDPData,
+        new_model::JuMP.AbstractModel,
+        old_model::JuMP.AbstractModel
+    )::GDPData
+
+Extend `JuMP.copy_extension_data` to initialize an empty [`GDPData`](@ref) object 
+for the copied model. This is the first step in the model copying process and is 
+automatically called by `JuMP.copy_model`. The actual GDP data (logical variables, 
+disjunctions, etc.) is copied separately via [`copy_gdp_data`](@ref).
+"""
 function JuMP.copy_extension_data(
     data::GDPData{M, V, C, T},
     new_model::JuMP.AbstractModel,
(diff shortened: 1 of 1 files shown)


dnguyen227:
@pulsipher this is ready for review.

dnguyen227:
found issue

src/utilities.jl
```diff
@@ -6,4 +6,122 @@ function _copy_model(
...
+        new_gdp.variable_bounds[var_map[v]] = bounds
+    end
+
+    new_gdp.solution_method = old_gdp.solution_method
+    new_gdp.ready_to_optimize = old_gdp.ready_to_optimize
+
+    return lv_map
+end
+
+function copy_model_and_gdp_data(model::M) where {M <: JuMP.AbstractModel}
```
pulsipher: Let's call this `copy_gdp_model`

PR #132, opened 2026-01-08
Fix MBM Implementation.

dnguyen227:
This PR fixes the MBM implementation to use the tightest M values possible. This means that each disjunct constraint has it's own M value associated with any other disjunct. Originally only disjuncts has corresponding M values. Infeasible and negative M values are also now used to simplify the overall problem upon reformulation (infeasible problems deactivate disjuncts and sets of negative or 0 M values reveal global constraints).

**EXAMPLE**

  Given a disjunction with:
  - **Y[1]**: `x <= 2` and `x >= 1`
  - **Y[2]**: `0 <= x <= 55`

  When reformulating Y[1]'s constraints with respect to Y[2]'s feasible region (0 ≤ x ≤ 55):

  - `x <= 2`: max(x - 2) → M = **53** (at x = 55)
  - `x >= 1`: max(1 - x) → M = **1** (at x = 0)

  **Before (max M of possible constraints):**
  x - 53·Y[2] <= 2
  x + 53·Y[2] >= 1   ← Both use M = max(53, 1) = 53

  **After (per-constraint):**
  x - 53·Y[2] <= 2   ← Uses its own M = 53
  x +  1·Y[2] >= 1   ← Uses its own M = 1 (tighter)

The code is largely the same. The change made is `maximum(_maximize_M(...))` is now just  `_maximize_M(...)`, and order of for loops was adjusted to reflect this.

Relevant tests were added to check for these unique M values in addition to verbose comments.

### DIFF
--- a/src/datatypes.jl
+++ b/src/datatypes.jl
@@ -392,13 +392,17 @@ mutable struct _MBM{O, T, M <: JuMP.AbstractModel} <: AbstractReformulationMetho
     default_M::T
     conlvref::Vector{LogicalVariableRef{M}}
     deactivated::Set{LogicalVariableRef{M}}
+    # Stored submodels: indicator => (submodel, var_map)
+    store::Dict{LogicalVariableRef{M}, Tuple{M, Dict}}
 
     function _MBM(method::MBM{O, T}, model::M) where {O, T, M <: JuMP.AbstractModel}
-        new{O, T, M}(method.optimizer,
+        new{O, T, M}(
+            method.optimizer,
             Dict{LogicalVariableRef{M}, Union{T, Vector{T}}}(),
             method.default_M,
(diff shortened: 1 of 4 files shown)


src/datatypes.jl
```diff
@@ -388,15 +388,15 @@ end
 
 mutable struct _MBM{O, T, M <: JuMP.AbstractModel} <: AbstractReformulationMethod
     optimizer::O
-    M::Dict{LogicalVariableRef{M}, T}                        
-    default_M::T                            
-    conlvref::Vector{LogicalVariableRef{M}}                  
+    M::Dict{LogicalVariableRef{M}, Union{T, Vector{T}}}
```
dnguyen227, line 391: Now storing M values for vector constraints as vectors to reflect unique values in each of rows.

PR #133, opened 2026-03-23
MBM update (preparation for InfGDP)

dnguyen227:
This PR refactors the MBM implementation to use a unified submodel infrastructure (GDPSubmodel) with cached submodels and infeasible disjunct detection. The
  _mini_model + _constraint_to_objective pattern is replaced by create_submodel + prepare_objectives + _raw_M, which separates submodel construction from objective evaluation and enables extension by InfiniteOpt.

  EXAMPLE

  Given a disjunction with:
  - Y[1]: x <= 2 and x >= 1
  - Y[2]: 0 <= x <= 55

  When reformulating Y[1]'s constraints with respect to Y[2]'s feasible region (0 ≤ x ≤ 55):

  - x <= 2 : max(x - 2) → M = 53 (at x = 55)
  - x >= 1 : max(1 - x) → M = 1 (at x = 0)

  Before (single M per disjunct pair, no caching):
  Each call to _mini_model creates a new submodel from scratch. M values are computed via _constraint_to_objective which sets the objective directly. No
  infeasible disjunct detection — unbounded subproblems silently use default_M.

  After (per-constraint M, cached submodels, infeasibility detection):
  - create_submodel builds and caches submodels in method.store (one per disjunct)
  - prepare_objectives converts constraints to maximization objectives
  - _raw_M solves and handles infeasible (returns nothing), unbounded (uses default_M + clears NaN start values), and optimal (clamps to ≥ 0) cases
  - Infeasible disjuncts are detected and deactivated with a warning
  - Sets of all-zero M values are enforced globally (no binary variables needed)

  Key structural changes:
  - _mini_model → create_submodel + _raw_M (separated concerns)
  - _constraint_to_objective → prepare_objectives (public extension point, dispatch on LessThan/GreaterThan)
  - _replace_variables_in_constraint now uses AbstractDict to handle mixed Number/VariableRef maps (needed by InfiniteOpt extension where parameter functions
  evaluate to Numbers at support points)
  - _MBM struct gains deactivated::Set and store::Dict fields
  - GDPSubmodel struct added to datatypes.jl (shared by MBM and CP)

  Relevant tests were added to cover per-constraint M values, infeasible disjunct detection, global constraint detection, and unbounded subproblem fallback.

### DIFF
--- a/src/datatypes.jl
+++ b/src/datatypes.jl
@@ -472,9 +472,22 @@ end
 #                              GDP SUBMODEL
 ################################################################################
 
-# Unified submodel wrapper for MBM and cutting planes.
-# Holds a flat JuMP model, ordered decision variables,
-# and a forward map (orig var → submodel vars).
+"""
+    GDPSubmodel{M, V, W}
+
+A unified submodel wrapper used by MBM and cutting plane
+reformulations. It encapsulates a flat JuMP optimization
+submodel built from a single disjunct's feasible region,
+along with mappings back to the original model's variables.
+
+## Fields
+- `model::M`: The JuMP submodel representing a disjunct's
+   feasible region (constraints and variable bounds).
+- `dec_vars::Vector{V}`: Ordered decision variables in
+   the submodel, matching the original model's ordering.
+- `fwd::Dict{V, Vector{W}}`: Forward map from original
+   model variables to their submodel counterparts.
+"""
 struct GDPSubmodel{M <: JuMP.AbstractModel,
                    V <: JuMP.AbstractVariableRef,
                    W <: JuMP.AbstractVariableRef}
--- a/src/datatypes.jl
+++ b/src/datatypes.jl
@@ -388,15 +388,22 @@ end
 
 mutable struct _MBM{O, T, M <: JuMP.AbstractModel} <: AbstractReformulationMethod
     optimizer::O
-    M::Dict{LogicalVariableRef{M}, T}                        
-    default_M::T                            
-    conlvref::Vector{LogicalVariableRef{M}}                  
+    M::Dict{LogicalVariableRef{M}, Any}
+    default_M::T
+    conlvref::Vector{LogicalVariableRef{M}}
+    deactivated::Set{LogicalVariableRef{M}}
+    # Stored submodels: indicator => GDPSubmodel.
+    # Typed Any so extensions can store different types.
+    store::Dict{LogicalVariableRef{M}, Any}
 
     function _MBM(method::MBM{O, T}, model::M) where {O, T, M <: JuMP.AbstractModel}
-        new{O, T, M}(method.optimizer,
-            Dict{LogicalVariableRef{M}, T}(), 
+        new{O, T, M}(
+            method.optimizer,
+            Dict{LogicalVariableRef{M}, Any}(),
             method.default_M,
-            Vector{LogicalVariableRef{M}}()                               
+            Vector{LogicalVariableRef{M}}(),
+            Set{LogicalVariableRef{M}}(),
+            Dict{LogicalVariableRef{M}, Any}()
         )
     end
 end
@@ -444,21 +451,36 @@ A type for using the cutting planes approach for disjunctive constraints.
 method to use after cutting planes (default = `BigM()`).
 - `M_value::Float64`: Big-M value to use in the final reformulation (default = `1e9`).
 """
-struct cutting_planes{O} <: AbstractReformulationMethod
+struct cutting_planes{O, T} <: AbstractReformulationMethod
     optimizer::O;
     max_iter::Int
-    seperation_tolerance::Float64
+    seperation_tolerance::T
     final_reform_method::AbstractReformulationMethod
-    M_value::Float64
+    M_value::T
     function cutting_planes(
-        optimizer::O; 
-        max_iter::Int = 3, 
-        seperation_tolerance::Float64 = 1e-6, 
-        final_reform_method = BigM(), 
-        M_value::Float64 = 1e9
-    ) where {O}
-        new{O}(optimizer, max_iter, seperation_tolerance, final_reform_method, M_value)
+        optimizer::O;
+        max_iter::Int = 3,
+        seperation_tolerance::T = 1e-6,
+        final_reform_method = BigM(),
+        M_value::T = 1e9
+    ) where {O, T}
+        new{O, T}(optimizer, max_iter, seperation_tolerance, final_reform_method, M_value)
     end
+end
+
+################################################################################
+#                              GDP SUBMODEL
+################################################################################
+
+# Unified submodel wrapper for MBM and cutting planes.
+# Holds a flat JuMP model, ordered decision variables,
+# and a forward map (orig var → submodel vars).
+struct GDPSubmodel{M <: JuMP.AbstractModel,
+                   V <: JuMP.AbstractVariableRef,
+                   W <: JuMP.AbstractVariableRef}
+    model::M
+    dec_vars::Vector{V}
+    fwd::Dict{V, Vector{W}}
 end
 
 """
@@ -635,9 +657,9 @@ end
 """
     VariableProperties(expr)::VariableProperties
 
-Creates a `VariableProperties` object with blank variable info (no bounds, not fixed, 
-not binary/integer) from an expression. The `expr` argument is provided for 
-extensions to infer additional properties (e.g., parameter dependencies in InfiniteOpt).
+Creates a `VariableProperties` object with blank variable info (no bounds, not fixed,
+not binary/integer) from an expression. The `expr` argument is provided for
+extensions to infer additional properties.
 
 ## Arguments
 - `expr`: Expression for extensions to extract metadata from
--- a/src/datatypes.jl
+++ b/src/datatypes.jl
@@ -391,7 +391,6 @@ mutable struct _MBM{O, T, M <: JuMP.AbstractModel} <: AbstractReformulationMetho
     M::Dict{LogicalVariableRef{M}, Any}
     default_M::T
     subproblem_indicators::Vector{LogicalVariableRef{M}}
-    deactivated::Set{LogicalVariableRef{M}}
     # Cached submodels: indicator => GDPSubmodel.
     # Typed Any so extensions can store different types.
     model_cache::Dict{LogicalVariableRef{M}, Any}
@@ -402,7 +401,6 @@ mutable struct _MBM{O, T, M <: JuMP.AbstractModel} <: AbstractReformulationMetho
             Dict{LogicalVariableRef{M}, Any}(),
             method.default_M,
             Vector{LogicalVariableRef{M}}(),
-            Set{LogicalVariableRef{M}}(),
             Dict{LogicalVariableRef{M}, Any}()
         )
     end
--- a/src/datatypes.jl
+++ b/src/datatypes.jl
@@ -439,17 +439,17 @@ mutable struct _Hull{V <: JuMP.AbstractVariableRef, T} <: AbstractReformulationM
 end
 
 """
-    cutting_planes{O} <: AbstractReformulationMethod
+    cutting_planes{O,T} <: AbstractReformulationMethod
 
 A type for using the cutting planes approach for disjunctive constraints.
 
 **Fields**
 - `optimizer::O`: Optimizer to use when solving mini-models (required).
 - `max_iter::Int`: Number of iterations (default = `3`).
-- `seperation_tolerance::Float64`: Tolerance for the separation problem (default = `1e-6`).
+- `seperation_tolerance::T`: Tolerance for the separation problem (default = `1e-6`).
 - `final_reform_method::AbstractReformulationMethod`: Final reformulation 
 method to use after cutting planes (default = `BigM()`).
-- `M_value::Float64`: Big-M value to use in the final reformulation (default = `1e9`).
+- `M_value::T`: Big-M value to use in the final reformulation (default = `1e9`).
 """
 struct cutting_planes{O, T} <: AbstractReformulationMethod
     optimizer::O;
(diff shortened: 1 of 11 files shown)


pulsipher:
@dnguyen227 I am open to arguments otherwise, but I am not sure that I see the utility of deactivating disjuncts. If a constraint is infeasible, we should probably just throw an error. This would simplify things and I am not sure deleting bad disjuncts leads to a useful formulation.

src/datatypes.jl
```diff
@@ -388,15 +388,22 @@ end
 
 mutable struct _MBM{O, T, M <: JuMP.AbstractModel} <: AbstractReformulationMethod
     optimizer::O
-    M::Dict{LogicalVariableRef{M}, T}                        
-    default_M::T                            
-    conlvref::Vector{LogicalVariableRef{M}}                  
+    M::Dict{LogicalVariableRef{M}, Any}
+    default_M::T
+    conlvref::Vector{LogicalVariableRef{M}}
```
pulsipher: `conlvref` is not a very clear name, perhaps refine the name

src/datatypes.jl
```diff
@@ -388,15 +388,22 @@ end
...
+    deactivated::Set{LogicalVariableRef{M}}
+    # Stored submodels: indicator => GDPSubmodel.
+    # Typed Any so extensions can store different types.
+    store::Dict{LogicalVariableRef{M}, Any}
```
pulsipher: Again `store` is not a very clear name

src/datatypes.jl
```diff
@@ -432,35 +439,63 @@ mutable struct _Hull{V <: JuMP.AbstractVariableRef, T} <: AbstractReformulationM
...
 - `max_iter::Int`: Number of iterations (default = `3`).
-- `seperation_tolerance::Float64`: Tolerance for the separation problem (default = `1e-6`).
+- `seperation_tolerance::T`: Tolerance for the separation problem (default = `1e-6`).
 - `final_reform_method::AbstractReformulationMethod`: Final reformulation 
 method to use after cutting planes (default = `BigM()`).
-- `M_value::Float64`: Big-M value to use in the final reformulation (default = `1e9`).
+- `M_value::T`: Big-M value to use in the final reformulation (default = `1e9`).
 """
-struct cutting_planes{O} <: AbstractReformulationMethod
+struct cutting_planes{O, T} <: AbstractReformulationMethod
```
pulsipher: I missed this before, types need to be upper case CammelCase

src/mbm.jl
```diff
@@ -1,477 +1,575 @@
...
+    model::JuMP.AbstractModel,
     disj::Disjunction,
     method::MBM
-)
+    )
     mbm = _MBM(method, model)
-    ref_cons = Vector{JuMP.AbstractConstraint}() 
+    disjunct_cons = Dict{LogicalVariableRef, Vector{JuMP.AbstractConstraint}}()
+    for d in disj.indicators
+        d in mbm.deactivated && continue
```
pulsipher: Ya, I am not sure we need this

src/mbm.jl
```diff
@@ -1,477 +1,575 @@
...
-    model::JuMP.AbstractModel, 
-    ref_cons::Vector{JuMP.AbstractConstraint}, 
-    lvref::LogicalVariableRef,
-    method::_MBM
-) 
-    
-    empty!(method.M)
+    model::JuMP.AbstractModel,
+    ref_cons::Vector{JuMP.AbstractConstraint},
+    lvref::LogicalVariableRef, method::_MBM
```
pulsipher: `method` should be on the next line

src/mbm.jl
```diff
@@ -1,477 +1,575 @@
...
     end
     return ref_cons
 end
 
 function reformulate_disjunct_constraint(
-    model::JuMP.AbstractModel,  
-    con::Disjunction, 
-    bconref::Union{Dict{<:LogicalVariableRef,<:JuMP.AbstractVariableRef},
-                   Dict{<:LogicalVariableRef,<:JuMP.GenericAffExpr}},
+    model::JuMP.AbstractModel, con::Disjunction,
```
pulsipher: `con` should be on next line

src/mbm.jl
```diff
@@ -1,477 +1,575 @@
...
-        ) for i in 1:objective.set.dimension
     )
+    M_vals = typeof(method.default_M)[]
+    for obj_expr in objectives
+        JuMP.@objective(sub.model, Max, obj_expr)
+        # Clear start values before each solve to prevent NaN
+        # residue from a previous non-feasible solve
+        for v in JuMP.all_variables(sub.model)
+            JuMP.set_start_value(v, nothing)
+        end
```
pulsipher: I don't quite follow why this is needed

src/mbm.jl
```diff
@@ -1,477 +1,575 @@
...
+        JuMP.optimize!(sub.model)
+        if JuMP.termination_status(sub.model) == _MOI.INFEASIBLE
+            return nothing
+        elseif !JuMP.is_solved_and_feasible(sub.model)
+            push!(M_vals, method.default_M)
+        else
+            push!(M_vals, max(
+                JuMP.objective_value(sub.model),
+                zero(method.default_M))
+                )
```
pulsipher: What type of solution is assumed here?

src/mbm.jl
```diff
@@ -1,477 +1,575 @@
...
+during constraint reformulation. The base method returns the single
+element from a length-1 vector. Extensions (e.g., InfiniteOpt) override
+to aggregate across multiple support points (e.g., interpolating K
+per-support M values into a parameter function).
+"""
+function condense_values(
+    ::JuMP.AbstractModel,
+    vals::AbstractVector
+    )
+    return vals[1]
```
pulsipher: probably safer to use `only(vals)`

src/mbm.jl
```diff
@@ -1,477 +1,575 @@
...
+"""
+    condense_values(model, vals::AbstractVector)
+
+Reduce a vector of raw M values from `_raw_M` to the final form used
+during constraint reformulation. The base method returns the single
+element from a length-1 vector. Extensions (e.g., InfiniteOpt) override
+to aggregate across multiple support points (e.g., interpolating K
+per-support M values into a parameter function).
+"""
+function condense_values(
```
pulsipher: should use more specific name like `aggregate_M_values`

src/datatypes.jl
pulsipher: something like `model_cache` would be more clear

src/mbm.jl
```diff
@@ -1,477 +1,575 @@
...
-    return _mini_model(model, objective, constraints, method)
+    ) where {T, S <: _MOI.Interval}
+    sub = _get_submodel(model, constraints, method)
+    set_values = _set_values(objective.set)
+    ge_obj = JuMP.ScalarConstraint(objective.func,
+        MOI.GreaterThan(set_values[1]))
+    le_obj = JuMP.ScalarConstraint(objective.func,
+        MOI.LessThan(set_values[2]))
+    raw_lower = _raw_M(sub,prepare_objectives(model, ge_obj, sub),method)
+    raw_upper = _raw_M(sub,prepare_objectives(model, le_obj, sub),method)
```
pulsipher: Missing spaces. This happens across the instances of this method

src/mbm.jl
```diff
@@ -1,477 +1,575 @@
...
-    constraints::Vector{<:DisjunctConstraintRef}, 
+"""
+    create_submodel(model, constraints, method::_MBM)
+
+Build a `GDPSubmodel` representing a disjunct's feasible region for
+MBM subproblem solves. Copies the model's decision variables and adds
+the given disjunct constraints. Submodels are cached in `method.store`
+by indicator.
+"""
+function create_submodel(
```
pulsipher: is this specific to only mbm? If so, we should make the naming more clear. Otherwise, how might this be used for other methods?

pulsipher (changes requested):

src/mbm.jl
dnguyen227: Let me know if this is too early in development but I've considered that LOA can use this function for the NLP subproblems.

To do the MBM subproblem I need to pass a feasible region described by one disjunct.
To do the LOA NLP subproblem I need to pass a feasible region described by a selected combination of disjuncts.

pulsipher: Ok, let's just come up with a better name and make the docstring more general

src/mbm.jl
dnguyen227: Sorry could you clarify the question?

We're expecting positive solutions to use as our M values. However it's possible to have negative values indicating that the constraint is constraint is always satisfied inside the disjunct's region. M_vals in general is a vector of scalar values though.

dnguyen227:
> @dnguyen227 I am open to arguments otherwise, but I am not sure that I see the utility of deactivating disjuncts. If a constraint is infeasible, we should probably just throw an error. This would simplify things and I am not sure deleting bad disjuncts leads to a useful formulation.

I overlooked this in my original implementation, below is the paper's call to remove infeasible disjuncts.

> "... if (2) is infeasible, then the term ki′ is infeasible and can be removed from the original problem.**"

Where (2) is the subproblem solving for M, and ki' is the disjunct term.

With regard to the utility of actually implementing this. I'm assuming that for the infinite-dimensional case, disjuncts can become infeasible at certain support points while being feasible in others, and this is would be worth the hassle of having code to deactivate disjuncts (at specific support points). 

Let me know your thoughts.

src/mbm.jl
pulsipher: This is part of a larger conditional statement based on optimality/feasibility and is under the `else` part which likely has implicit assumptions.

pulsipher:
> > @dnguyen227 I am open to arguments otherwise, but I am not sure that I see the utility of deactivating disjuncts. If a constraint is infeasible, we should probably just throw an error. This would simplify things and I am not sure deleting bad disjuncts leads to a useful formulation.
> 
> I overlooked this in my original implementation, below is the paper's call to remove infeasible disjuncts.
> 
> > "... if (2) is infeasible, then the term ki′ is infeasible and can be removed from the original problem.**"
> 
> Where (2) is the subproblem solving for M, and ki' is the disjunct term.
> 
> With regard to the utility of actually implementing this. I'm assuming that for the infinite-dimensional case, disjuncts can become infeasible at certain support points while being feasible in others, and this is would be worth the hassle of having code to deactivate disjuncts (at specific support points).
> 
> Let me know your thoughts.

If I understand correctly, an infeasible problem means the constraints in a disjunct are fundamentally infeasible regardless of the global constraints and other disjuncts. If this is true, then seems like modelling it as such is a fundamental modelling mistake.

src/mbm.jl
dnguyen227: I've updated the structure to make it more logical (this depends on the convo with deactivation). Now we test the following:

Did we get a solution and its feasible? -> If yes push the M value.
If not, did we get a solution but its infeasible? -> If yes then return nothing (to signal deactivation or throw error)
If none of the above, lets safely default to an M value as this cant hurt or improve the problem as long as it's sufficiently large enough.

pulsipher: ok

dnguyen227:
> > > @dnguyen227 I am open to arguments otherwise, but I am not sure that I see the utility of deactivating disjuncts. If a constraint is infeasible, we should probably just throw an error. This would simplify things and I am not sure deleting bad disjuncts leads to a useful formulation.
> > 
> > 
> > I overlooked this in my original implementation, below is the paper's call to remove infeasible disjuncts.
> > > "... if (2) is infeasible, then the term ki′ is infeasible and can be removed from the original problem.**"
> > 
> > 
> > Where (2) is the subproblem solving for M, and ki' is the disjunct term.
> > With regard to the utility of actually implementing this. I'm assuming that for the infinite-dimensional case, disjuncts can become infeasible at certain support points while being feasible in others, and this is would be worth the hassle of having code to deactivate disjuncts (at specific support points).
> > Let me know your thoughts.
> 
> If I understand correctly, an infeasible problem means the constraints in a disjunct are fundamentally infeasible regardless of the global constraints and other disjuncts. If this is true, then seems like modelling it as such is a fundamental modelling mistake.

Fair enough. I've removed the deactivation logic and replaced it with an informative error.

src/datatypes.jl
```diff
@@ -432,35 +437,63 @@ mutable struct _Hull{V <: JuMP.AbstractVariableRef, T} <: AbstractReformulationM
...
+   the submodel, matching the original model's ordering.
+- `fwd::Dict{V, Vector{W}}`: Forward map from original
+   model variables to their submodel counterparts.
+"""
+struct GDPSubmodel{M <: JuMP.AbstractModel,
+                   V <: JuMP.AbstractVariableRef,
+                   W <: JuMP.AbstractVariableRef}
+    model::M
+    dec_vars::Vector{V}
+    fwd::Dict{V, Vector{W}}
```
pulsipher: let's use the word map in the name here

src/mbm.jl
```diff
@@ -1,477 +1,562 @@
...
 ################################################################################
-# Dispatches over constraint types to reformulate into >= or <= 
-# in order to solve the mini-model
-function _maximize_M(
-    model::JuMP.AbstractModel, 
-    objective::JuMP.VectorConstraint{T, S, R}, 
-    constraints::Vector{<:DisjunctConstraintRef}, 
+
+"""
+    prepare_objectives(model, obj::ScalarConstraint, sub::GDPSubmodel)
```
pulsipher: Maybe we should have a more precise name like `prepare_max_M_objective`. We can probably make it singular as well and have InfiniteOpt return a singular vector and have JuMP return a single expression (no vector) and dispatch accordingly in the appropriate methods below.

test/constraints/mbm.jl
```diff
@@ -1,9 +1,55 @@
...
+    result2 = DP._replace_variables_in_constraint(quad1, map2)
+    @test result2.aff.terms[y] ≈ 2.0
+
+    # rb Number, ra VariableRef (line 574)
+    map3 = Dict{VariableRef, Any}(x[1] => y, x[2] => 3.0)
+    result3 = DP._replace_variables_in_constraint(quad1, map3)
+    @test result3.aff.terms[y] ≈ 3.0
 end
 
 function test__replace_variables_in_constraint()
```
pulsipher: This has an extra underscore

pulsipher (changes requested):
Nearly there, just a few minor suggestions.

pulsipher (approved):

PR #134, opened 2026-04-02
Cutting planes update (Prep for InfGDP)

dnguyen227:
This PR refactors the cutting planes implementation and provides extension points for InfiniteOpt integration.

### DIFF
--- a/src/cuttingplanes.jl
+++ b/src/cuttingplanes.jl
@@ -1,149 +1,138 @@
-function reformulate_model(
-    model::JuMP.AbstractModel, 
-    method::CuttingPlanes
-    ) 
-    _clear_reformulations(model)
-    var_type = JuMP.variable_ref_type(model)
-    obj = objective_function(model)
-    sense = objective_sense(model)
+################################################################################
+#                       FUNCTIONS FOR LOOP
+################################################################################
 
-    #Creation of seperation (SEP) and relaxed big M model (rBM).
-    SEP, sep_ref_map, _ = copy_gdp_model(model)
-    rBM, rBM_ref_map, _ = copy_gdp_model(model)
-    reformulate_model(rBM, BigM(method.M_value))
-    reformulate_model(SEP, Hull())
-    main_to_SEP_map = Dict(v => sep_ref_map[v] for v in collect_all_vars(model))
-    main_to_rBM_map = Dict(v => rBM_ref_map[v] for v in collect_all_vars(model))
-    JuMP.set_optimizer(SEP, method.optimizer)
-    JuMP.set_optimizer(rBM, method.optimizer)
-    JuMP.set_silent(rBM)
-    JuMP.set_silent(SEP)
-    JuMP.relax_integrality(rBM)
-    JuMP.relax_integrality(SEP)
-    JuMP.@objective(rBM, sense, 
-        _replace_variables_in_constraint(obj, main_to_rBM_map)
+# Collect decision variables for cutting planes. Extensions may override to
+# customize variable collection.
+function collect_cutting_planes_vars(model::JuMP.AbstractModel)
+    return collect_all_vars(model)
+end
+
+# Extract solution from a solved model (in-place). Extensions
+# override for models where values live on a backend.
+function extract_solution(model::JuMP.AbstractModel)
+    dvars = collect_cutting_planes_vars(model)
+    V = eltype(dvars)
+    T = JuMP.value_type(typeof(model))
+    return Dict{V, Vector{T}}(
+        v => [JuMP.value(v)] for v in dvars)
+end
+
+# Extract solution from a GDPSubmodel (SEP path).
+function extract_solution(sub::GDPSubmodel)
+    V = eltype(sub.decision_vars)
+    T = JuMP.value_type(typeof(sub.model))
+    sol = Dict{V, Vector{T}}()
+    for var in sub.decision_vars
+        sol[var] = JuMP.value.(sub.fwd_map[var])
+    end
+    return sol
+end
+
+# Set quadratic separation objective: min Σ (x_k - rBM_k)².
+function _set_separation_objective(
+    sub::GDPSubmodel,
+    rBM_sol::Dict{<:JuMP.AbstractVariableRef, <:Vector{<:Number}}
     )
-    #Mapping of variables between models.
-    rBM_to_SEP_map = Dict{var_type, var_type}()
-    SEP_to_rBM_map = Dict{var_type, var_type}()
-    for (var, rBM_var) in main_to_rBM_map
-        SEP_var = main_to_SEP_map[var]
-        rBM_to_SEP_map[rBM_var] = SEP_var
-        SEP_to_rBM_map[SEP_var] = rBM_var
-    end
-    
-    #Main cutting planes loop.
-    i = 1
-    sep_obj = Inf
-    while i <= method.max_iter && sep_obj > method.seperation_tolerance
-        rBM_sol = _solve_rBM(rBM)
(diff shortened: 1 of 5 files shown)


src/cuttingplanes.jl
```diff
@@ -1,149 +1,176 @@
-function reformulate_model(
-    model::JuMP.AbstractModel, 
+################################################################################
+#                       FUNCTIONS FOR LOOP
+################################################################################
+
+# Collect decision variables for cutting planes. Extensions may override to
+# customize variable collection.
+function collect_cp_vars(model::JuMP.AbstractModel)
```
pulsipher: We should avoid acronyms in names like this, use `cutting_planes` instead of `cp`. Makes the code much more clear down the road.

src/cuttingplanes.jl
```diff
@@ -1,149 +1,176 @@
...
+    return collect_all_vars(model)
+end
+
+# Build one cutting planes subproblem (SEP). dec_vars is the shared key space
+# (collected once from the clean model). Extensions may override for custom
+# subproblem construction.
+function build_cp_subproblem(
```
pulsipher: Again, avoid using `cp` abbreviation

src/cuttingplanes.jl
```diff
@@ -1,149 +1,176 @@
...
+    model::JuMP.AbstractModel,
+    dec_vars::AbstractVector,
```
pulsipher: using `decision` instead of `dec`. Can we also be more specific with the vector type?

src/cuttingplanes.jl
```diff
@@ -1,149 +1,176 @@
...
+    JuMP.set_optimizer(sub.model, method.optimizer)
+    JuMP.set_silent(sub.model)
+    return sub
+end
 
-    #Final reformulation with added cutting planes.
-    reformulate_model(model, method.final_reform_method)
-    return
+# Set up the rBM (relaxed Big-M) subproblem.
+function setup_rbm(
```
pulsipher: Again, let's not use acronyms like this. Also, would it not be possible to make a more general function that simply makes a relaxed model with an arbitrary input method? It seems like that could be used elsewhere.

src/cuttingplanes.jl
```diff
@@ -1,149 +1,176 @@
...
-    rBM_to_SEP_map::Dict{<:JuMP.AbstractVariableRef,<:JuMP.AbstractVariableRef} 
-    ) where {M <: JuMP.AbstractModel, T <: Number}
-
-    SEP_vars = [rBM_to_SEP_map[rBM_var] for rBM_var in collect_all_vars(rBM)]
-
-    #Modified objective function for SEP.
-    obj_expr = sum(
-        (SEP_var - rBM_sol[SEP_to_rBM_map[SEP_var]])^2 for SEP_var in SEP_vars
+# Set quadratic separation objective: min Σ (x_k - rBM_k)².
+function _set_sep_objective(
```
pulsipher: again, not clear what `sep` is

src/cuttingplanes.jl
pulsipher: Can you clarify this method. It modifies the main model in place and creates a submodel. That seems to be doing two different things. To reformulate the model in place and relax it probably does not need to be in its own function anyways.

pulsipher (changes requested):
See below. Please avoid abbreviations unless they are universally known (e.g., CPU).

src/cuttingplanes.jl
dnguyen227: fyi I've renamed this function to reformulate_and_relax and moved it to utilities.jl along with other functions similar because they're pretty general when I pondered on it.

The reason for making an function to do this is to have an extension point for InfiniteGDP. 

In the finite case we're able to have just the 2 models (manipulating integrality to get the rBM and recover the original model). **For the infinite case we have 3 models - the original model (infinite), the transcribed rBM problem, and the transcribed separation problem.** This extension point allowed me to dispatch these few lines instead of the outer function.

I could've misunderstood our convo, but I thought we were trying to avoid calling optimize on the original model (infinite) because it needs to rebuild from scratch each time we add a constraint (or cut in this case)?

src/utilities.jl
```diff
@@ -8,6 +8,96 @@ function _copy_model(
...
+    copy_model_with_constraints(model, constraints, method)
+
+Create a new model with only the variables (and their bounds)
+from `model` and the selected `constraints`. Returns a
+`GDPSubmodel` wrapping the new model.
+"""
+function copy_model_with_constraints(
+    model::JuMP.AbstractModel,
+    constraints::Vector{<:DisjunctConstraintRef},
+    method::_MBM
```
pulsipher: It this specific only to `_MBM`

src/cuttingplanes.jl
pulsipher: Could you show code snippets of this workflow for a JuMP vs an InfiniteOpt model to make it more clear?

pulsipher (changes requested):

src/utilities.jl
dnguyen227: Good point, I've moved it back.

src/cuttingplanes.jl
dnguyen227: So without pasting the entire PR, here are snippets for the `relax_and_reformulate`, as well as how the main loop looks for your reference.

1. Shared main function (same for both cases, calls `copy_and_reformulate` and `reformulate_and_relax`):

```julia
  # From reformulate_model in cuttingplanes.jl
  separation = copy_and_reformulate(model, decision_vars, Hull(), method)
  JuMP.relax_integrality(separation.model)

  rBM, undo_relax = reformulate_and_relax(
      model, decision_vars, BigM(method.M_value), method)
```

2a. Finite `reformulate_and_relax` reformulates and relaxes the original model in-place. Only 2 models total (original + separation copy):

```julia
  # From utilities.jl
  function reformulate_and_relax(model, decision_vars, reform_method, method)
      reformulate_model(model, reform_method)  # modifies original in-place
      fwd_map = Dict(v => [v] for v in decision_vars)  # identity map
      sub = GDPSubmodel(model, decision_vars, fwd_map)
      undo_relax = JuMP.relax_integrality(model)
      return sub, undo_relax  # rBM.model === model
  end
```

2b. Infinite `reformulate_and_relax` creates a separate transcribed flat model instead of modifying in-place. 3 models total (original InfiniteModel + transcribed rBM + transcribed separation):
```julia
  # From ext/InfiniteDisjunctiveProgramming.jl
  function DP.reformulate_and_relax(model::InfiniteModel, decision_vars, reform_method, method)
      rBM = DP.copy_and_reformulate(model, decision_vars, reform_method, method)
      JuMP.relax_integrality(rBM.model)
      return rBM, nothing  # rBM.model !== model
  end
```

3. In the shared main loop of the main function, call optimize on rBM and seperation problems. We avoid calling optimize on the InfiniteOpt model here.
```julia
for iter in 1:method.max_iter
        # 1. Solve the relaxed Big-M (rBM) master problem for current iterate
        JuMP.optimize!(rBM.model, ignore_optimize_hook = true) 
        rBM_sol = _extract_solution(rBM)

        # 2. Solve the SEP (separation subproblem) using the rBM solution
        separation_obj, separation_sol = _solve_separation(separation, rBM_sol)

        # 3. Check convergence: separation_obj ≈ 0 means rBM solution is already
        # Hull-feasible, so no separating cut exists
        if separation_obj <= method.seperation_tolerance
            break
        end

        # 4. Add separating cut to rBM
        _add_cut(rBM, rBM_sol, separation_sol)
    end
```

The transcribed rBM model can be handled that same as in base package. Additional logic for cuts onto the infinite model are made in the next PR.

pulsipher: I think we can simply and improve this, let's discuss this in person

pulsipher (approved):

PR #135, opened 2026-04-20
Extensions to MBM and cutting planes techniques.

dnguyen227:
This PR makes changes to extend MBM and cutting planes techniques.  

## MBM
  - `copy_model_with_constraints(::InfiniteModel, ...)`
  - `prepare_max_M_objective(::InfiniteModel, ...)` (LessThan + GreaterThan)
  - `raw_M(sub, ::Vector{<:AbstractJuMPScalar}, ::_MBM)`

  ## Cutting Planes
  - `copy_and_reformulate(::InfiniteModel, ..., ::CuttingPlanes)`
  - `extract_solution(::InfiniteModel)`
  - `add_cut(::InfiniteModel, ...)`

### DIFF
--- a/ext/InfiniteDisjunctiveProgramming.jl
+++ b/ext/InfiniteDisjunctiveProgramming.jl
@@ -303,10 +303,15 @@ function DP.prepare_max_M_objective(
     return InfiniteOpt.transformation_expression(obj.set.lower - mini_expr)
 end
 
+# Real dispatch: pure-parameter slacks collapse to a constant at that
+# support. Mirrors the max(value, 0) semantics of the scalar base.
+DP.raw_M(::DP.GDPSubmodel, obj::Real, method::DP._MBM) =
+    max(obj, zero(method.default_M))
+
 # Solve the submodel for a vector of objectives (one per support point).
-# Elements may be Real when the slack is pure-parameter at that support;
-# JuMP's @objective accepts Real and the solver treats it as a constant
-# objective, which gives the correct M value for that slice.
+# Clears start values before each solve (Gurobi refuses NaN warmstarts
+# that can linger from a prior unbounded solve) and delegates each
+# element to the scalar base `raw_M` above.
 function DP.raw_M(
     sub::DP.GDPSubmodel,
     objectives::Vector{<:Union{Real, JuMP.AbstractJuMPScalar}},
@@ -315,18 +320,9 @@ function DP.raw_M(
     M_vals = typeof(method.default_M)[]
     for obj_expr in objectives
         JuMP.set_start_value.(JuMP.all_variables(sub.model), nothing)
-        JuMP.@objective(sub.model, Max, obj_expr)
-        JuMP.optimize!(sub.model)
-        if JuMP.is_solved_and_feasible(sub.model)
-            push!(M_vals, max(
-                JuMP.objective_value(sub.model),
-                zero(method.default_M)))
-        elseif JuMP.termination_status(sub.model) ==
-                JuMP.MOI.INFEASIBLE
-            return nothing
-        else
-            push!(M_vals, method.default_M)
-        end
+        m = DP.raw_M(sub, obj_expr, method)
+        m === nothing && return nothing
+        push!(M_vals, m)
     end
     model = sub.model.ext[:inf_mbm_main]
     # Condense per-support values: scalar if uniform, else pfunc.
--- a/ext/InfiniteDisjunctiveProgramming.jl
+++ b/ext/InfiniteDisjunctiveProgramming.jl
@@ -247,7 +247,6 @@ function DP.copy_model_with_constraints(
         Dict{InfiniteOpt.GeneralVariableRef, Vector{JuMP.VariableRef}}())
 end
 
-# Return one pointwise slack per support.
 function DP.prepare_max_M_objective(
     ::InfiniteOpt.InfiniteModel,
     obj::JuMP.ScalarConstraint{T, S},
--- a/ext/InfiniteDisjunctiveProgramming.jl
+++ b/ext/InfiniteDisjunctiveProgramming.jl
@@ -8,14 +8,17 @@ import DisjunctiveProgramming as DP
 #                                   MODEL
 ################################################################################
 function DP.InfiniteGDPModel(args...; kwargs...)
-    return DP.GDPModel{InfiniteOpt.InfiniteModel,
+    return DP.GDPModel{
+        InfiniteOpt.InfiniteModel,
         InfiniteOpt.GeneralVariableRef,
-        InfiniteOpt.InfOptConstraintRef}(args...; kwargs...)
+        InfiniteOpt.InfOptConstraintRef
+        }(args...; kwargs...)
 end
 
 function DP.collect_all_vars(model::InfiniteOpt.InfiniteModel)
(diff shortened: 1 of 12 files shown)


ext/InfiniteDisjunctiveProgramming.jl
```diff
@@ -171,17 +171,213 @@ function DP.disaggregate_expression(
...
-    error("The `MBM` method is not supported for `InfiniteModel`." *
-          "Please use `BigM`, `Hull`, `Indicator`, or `PSplit` instead.")
+
+# Build a mini InfiniteModel holding only the given disjunct constraints,
+# transcribe it, and return as a GDPSubmodel.
+function DP.copy_model_with_constraints(
+    model::InfiniteOpt.InfiniteModel,
+    constraints::Vector{<:DP.DisjunctConstraintRef},
+    method::DP._MBM
+    )
```
pulsipher, line 184: Based on the code I see below, there are a number of gotchas when it comes to copying `InfiniteModel`s, I think we should instead implement `JuMP.copy_model(model::InfiniteModel)` in InfiniteOpt first to then use here.

ext/InfiniteDisjunctiveProgramming.jl
```diff
@@ -171,17 +171,213 @@ function DP.disaggregate_expression(
...
+    end
+
+    # 6. Transcribe mini InfiniteModel
+    InfiniteOpt.build_transformation_backend!(mini)
+    transcribed = InfiniteOpt.transformation_model(mini)
+    JuMP.set_optimizer(transcribed, method.optimizer)
+    JuMP.set_silent(transcribed)
+    # Stash for prepare_max_M_objective / raw_M.
+    transcribed.ext[:inf_mbm_main] = model
+    transcribed.ext[:inf_mbm_ref_map] = ref_map
```
pulsipher: I don't understand why this stashing is needed here, but not for JuMP models.

ext/InfiniteDisjunctiveProgramming.jl
```diff
@@ -171,17 +171,213 @@ function DP.disaggregate_expression(
...
+    sub.model.ext[:inf_mbm_obj_expr] = obj.func
+    return InfiniteOpt.transformation_expression(mini_expr)
+end
+
+# Per-support solve, delegating to scalar base raw_M. Aggregated to a
+# scalar if uniform, else to a parameter function.
+function DP.raw_M(
+    sub::DP.GDPSubmodel,
+    objectives::AbstractArray{<:Union{JuMP.AbstractJuMPScalar, Real}},
+    method::DP._MBM
```
pulsipher: This is type piracy since none of these types are owned by InfiniteOpt

ext/InfiniteDisjunctiveProgramming.jl
```diff
@@ -171,17 +171,213 @@ function DP.disaggregate_expression(
...
+    )
+    M_vals = Array{typeof(method.default_M)}(undef, size(objectives))
+    for I in eachindex(objectives)
+        JuMP.set_start_value.(JuMP.all_variables(sub.model), nothing)
```
pulsipher: Why do you need to reset the start values? This could be problematic from some NLPs that rely on good start values.

ext/InfiniteDisjunctiveProgramming.jl
```diff
@@ -171,17 +171,213 @@ function DP.disaggregate_expression(
...
+        m = DP.raw_M(sub, objectives[I], method)
+        m === nothing && return nothing
+        M_vals[I] = m
+    end
+    all(==(first(M_vals)), M_vals) && return first(M_vals)
+    main = sub.model.ext[:inf_mbm_main]
+    expr = sub.model.ext[:inf_mbm_obj_expr]
```
pulsipher: Again, I don't quite understand why we are storing data this way

test/extensions/InfiniteDisjunctiveProgramming.jl
```diff
@@ -1,4 +1,4 @@
-using InfiniteOpt, HiGHS, Ipopt, Juniper
+using InfiniteOpt, HiGHS, Ipopt, Juniper, Interpolations
```
pulsipher: What happens if this is not imported? It seems strange for the user to import a package they do not directly use.

pulsipher (changes requested):
See the comments below. Also, we will need some documentation on how these methods are handled for InfiniteModels (there is some nuance).

ext/InfiniteDisjunctiveProgramming.jl
dnguyen227: Noted, I'll try to get some code going before Monday's meeting and I can try for a PR after we chat.

test/extensions/InfiniteDisjunctiveProgramming.jl
dnguyen227: When it's not imported I have all test sets failing (not limited to just raw_M where its called). 
`MethodError: no method matching InfiniteGDPModel()
    The function `InfiniteGDPModel` exists, but no method is defined for this combination of argument types.`

ext/InfiniteDisjunctiveProgramming.jl
dnguyen227: This is now addressed in the get_variable_info function. Based on my notes (and what I quickly verified just now), point variables are generated with a default start value of NaN. This was causing Gurobi to throw `ERROR: Gurobi Error 10003: Element 0 of a double array is Nan.`

Apologies as I intended to have this change in the current PR, it just didn't happen.

test/extensions/InfiniteDisjunctiveProgramming.jl
pulsipher: This is because Julia extensions cannot have additional dependencies not included in the base package. And due to some type piracy in Interpolations, I do not want it to be a core dependency of InfiniteOpt or DisjunctiveProgramming. If you only use it for linear interpolation of M-values, perhaps it makes since to have a lightweight native function instead. It might even make sense just to use constant interpolation.

ext/InfiniteDisjunctiveProgramming.jl
dnguyen227: Now dispatches on element inside GDPSubmodel.

ext/InfiniteDisjunctiveProgramming.jl
dnguyen227: Not at all needed. This will disappear.

ext/InfiniteDisjunctiveProgramming.jl
dnguyen227: Not at all needed. This will disappear.

dnguyen227:
@pulsipher Tests are failing with `MethodError: no method matching copy_model(::InfiniteModel)` (passing locally on my clone). Please advise on what I should do with setting compatible versions.

pulsipher:
> @pulsipher Tests are failing with `MethodError: no method matching copy_model(::InfiniteModel)` (passing locally on my clone). Please advise on what I should do with setting compatible versions.

This is b/c the master version was not be properly set up for testing on CI. However, InfiniteOpt v0.6.2 will be released in about 10 minutes, so just update the Project file to require 0.6.2 and you should be good to go.

pulsipher (approved):

PR #137, opened 2026-07-01
Implementing LOA and its infinite version.

dnguyen227:
Implements LOA with two main components.

- a primary NLP — the model reformulated by an inner_method (BigM, MBM, or Hull) with the disjunct binaries fixed at the current selection, and
- a master MILP that accumulates outer-approximation and no-good cuts until its bound meets the incumbent.

For the infinite implementation, the flow is as follows.
Starting from Infinite GDP,

1. Reformulate with the inner method (still infinite).
2. Transcribe once for the primary NLP.
3. Copy its variables + linear constraints this is the master (MILP) (I stay in this form for adding cuts and resolving).
4. Fix a combo on the transcribed NLP -> Solve.
5. Emit cuts into the transcribed master.


dnguyen227:
@pulsipher good for review.

src/datatypes.jl
```diff
@@ -417,8 +417,11 @@ constraints.
 """
 struct Hull{T} <: AbstractReformulationMethod
     value::T
-    function Hull(ϵ::T = 1e-6) where {T}
-        new{T}(ϵ)
+    # Internal: LOA installs a Dict here to collect the disaggregation
+    # map during reformulation; `nothing` for every other use.
+    disaggregation_map::Union{Nothing, AbstractDict}
```
pulsipher: This seems like something to avoid, is it not possible to only store it in `_Hull'?

src/loa.jl
```diff
@@ -0,0 +1,690 @@
...
+        # Overall cap: the final committed solve gets the budget left.
+        JuMP.set_time_limit_sec(nlp, max(0.0, overall_deadline - time()))
+    elseif isfinite(method.iteration_time_limit)
+        # Loop-only budget: restore so the final solve isn't crippled.
+        _restore_time_limit(nlp, original_time_limit)
+    end
+    _report_loa_gap(best_objective, best_result, master_bound, sense_token,
+        converged)
+    _set_solution_method(model, method)
+    _set_ready_to_optimize(model, true)
```
pulsipher: The problem I see with this workflow is that instead of reformulating, this actually just solves the problem as a solver. What happens if the user then calls `optimize!` afterward?

pulsipher (changes requested):
This took a lot of work. My main concern is that LOA is not really a reformulation method and we are trying to force it through the reformulation API, which has some gotchas. I suspect we will need a different workflow like how Pyomo.gdp does for methods like LOA.

PR #138, opened 2026-07-16
Fix tests for upcoming change in JuMP

odow:
JuMP is making a simplification to some non-linear expressions: https://github.com/jump-dev/JuMP.jl/pull/4188#issuecomment-4996925257


odow:
I'm also going to fix MOI so that we can do
```julia
@test isequal_canonical(
    simplify(ref[1].func),
    (x_z^2 / ((1 - ϵ) * zbin + ϵ)) - (5 * zbin),
)
```

pulsipher (approved):

pulsipher:
Thanks

PR #140, opened 2026-08-06
DisjunctionSet

dnguyen227:
This PR is a different way to implement #137. Instead #71 is implemented as well as an optimizer layer where the LOA algorithm will run. DP.jl own the DisjunctionSet that DA.jl unpacks to solve LOA.

This was tested on some problems from the SECQUOIA/gdplib, as well as the existing InfiniteGDP case studies we have (`InfiniteDisjunctiveProgramming.jl` works out of the box).


dnguyen227:
@pulsipher ready for review. This one is related to DisjunctiveAlgorithms.jl

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "DisjunctiveProgramming"
 uuid = "0d27d021-0159-4c7d-b4a7-9ccb5d9366cf"
 authors = ["hdavid16 <hperez16@gmail.com>"]
-version = "0.6.1"
+version = "0.6.2"
```
pulsipher, line 4: Changing the version doesn't belong in this PR

pulsipher (changes requested):
Looks good, just one minor update needed.

pulsipher (approved):

PR #141, opened 2026-08-07
Basic steps implementation

dnguyen227:
- apply_basic_step(model, [Y1, Y2, ...]) replaces the input disjunctions with their product: one new disjunct per combination, containing the union of the parents' constraints
- Global constraints can be intersected into every product disjunct
- Original logical vars Y are kept and linked via Y[i] in Exactly(w_slice) → reformulates to y_i = sum_j w_ij, so the model is equivalent.
- relax_products = true makes the product binaries continuous [0,1] (the linking equalities force them integral), so binaries don't multiply (Pyomo doesnt do this, its valid according to the lit).
- Right now, basic steps is destructive, as this all takes place before the user calls optimize and triggers the reformulation workflow.
- Also InfiniteGDP implementation (it doesnt seem to benefit solve times at all from my minimal testing).
```julia
model = GDPModel(HiGHS.Optimizer)
@variable(model, 0 <= x[1:2] <= 10)
@variable(model, Y[1:2], Logical)
@variable(model, Z[1:2], Logical)
@objective(model, Max, x[1] + x[2])
@constraint(model, x[1] <= 1, Disjunct(Y[1]))
@constraint(model, 2 <= x[1] <= 3, Disjunct(Y[2]))
@constraint(model, x[2] <= 1, Disjunct(Z[1]))
@constraint(model, 4 <= x[2] <= 5, Disjunct(Z[2]))
d1 = disjunction(model, Y)
d2 = disjunction(model, Z)
g = @constraint(model, x[1] + x[2] <= 6)

# replace d1 and d2 with their product disjunction (2x2 = 4 disjuncts),
# intersecting the global constraint g into every product disjunct
bs = apply_basic_step(model, [d1, d2], constraints = [g], name = "bs")

optimize!(model, gdp_method = Hull())   # tighter relaxation, same optimum
```


dnguyen227:
@pulsipher ready for review

src/basicsteps.jl
```diff
@@ -0,0 +1,309 @@
...
+    if con isa JuMP.ScalarConstraint &&
+        JuMP.jump_function(con) isa JuMP.AbstractVariableRef
+        error("Global constraint `$cref` constrains a single variable " *
+              "(e.g., a variable bound or integrality constraint) and " *
+              "cannot be intersected into a disjunction.")
+    end
+    supported = con isa JuMP.ScalarConstraint ||
+        (con isa JuMP.VectorConstraint && JuMP.moi_set(con) isa
+            Union{_MOI.Nonnegatives, _MOI.Nonpositives, _MOI.Zeros})
+    supported || error("Global constraint `$cref` with set " *
```
pulsipher, line 26: This won't play with the conic support we are adding

src/basicsteps.jl
```diff
@@ -0,0 +1,309 @@
...
+            if any(cref -> cref isa DisjunctionRef, crefs)
+                error("Basic steps do not support disjuncts that contain " *
+                      "nested disjunctions (indicator `$lvref`).")
+            end
+        end
+    end
+    all_indicators = [lv for d in disjunctions
+                      for lv in JuMP.constraint_object(d).indicators]
+    allunique(all_indicators) || error(
+        "The `disjunctions` for a basic step share indicator variables.")
```
pulsipher, line 71: Is this ever not true?

pulsipher (changes requested):

PR #142, opened 2026-08-24
Cutting-planes with quadrature weights for InfGDP

dnguyen227:
This PR adds the use of quadrature weights when making cuts on InfGDP problems. This makes `_CuttingPlanes` an inner type to store weights as an argument. While doing this, I brought some of the other things we keep track of into this as well to clean up the code. Theres a few paths I could've taken but think this is the cleanest at the sacrifice of holding the weights argument inside of the finite code (which is completely valid for users to change too at their own risk actually).


dnguyen227:
@pulsipher ready for review

ext/InfiniteDisjunctiveProgramming.jl
```diff
@@ -303,38 +303,130 @@ function DP.copy_and_reformulate(
...
+        if isempty(InfiniteOpt.parameter_refs(v))
+            weights[v] = [1.0]
+        else
+            tvars = vec(InfiniteOpt.transformation_variable(ref_map[v]))
+            weights[v] = [JuMP.coefficient(obj, tv) for tv in tvars]
+        end
+    end
+    return weights
+end
+
```
pulsipher, line 422: Rather than copy the model and transcribe it, you can extract the weights by directly interrogating the discrete measure data associated with the Measure object

pulsipher (changes requested):

pulsipher (approved):

